PR #12, opened 2019-08-25
add cmake configure step in readme for header only usage

XzoRit:
When I tried to use this library as header only it did not compile. The file config.hpp was missing. I figured out that if I run cmake as describe in the commit the according file got generated. So I thouhgt this should be reflected in the documentation.
If there is any other way to get the config.hpp generated please let me know.


arximboldi:
Thanks a lot for the PR!

PR #13, opened 2019-09-24
Add atria::funken based cursors

arximboldi:


arximboldi:
FYI @CJBussey :)

There are already some API changes compared to Atria, which I hope makes the system more inteligible. I would still like to improve the syntax for building derived cursors (the whole `xformed` shit), which is pretty bad now. I am gonna think a bit about it and explore the use of composable lenses.

Also, it seems like transducers are really slow to compile in their C++14 form. This is kind of driving me mad since it was a lot of effort to bring them to C++14...

PR #14, opened 2019-10-01
add missing includes to deps.hpp and event_loop/sdl.hpp

waybeforenow:
`deps.hpp` uses [`std::runtime_error`](https://en.cppreference.com/w/cpp/error/runtime_error) which is defined in `<stdexcept>`,

and `sdl.hpp` uses [`std::min`](https://en.cppreference.com/w/cpp/algorithm/min) which is defined in `<algorithm>`.

Both should be explicitly imported.

### DIFF
--- a/lager/deps.hpp
+++ b/lager/deps.hpp
@@ -24,6 +24,7 @@
 
 #include <functional>
 #include <optional>
+#include <stdexcept>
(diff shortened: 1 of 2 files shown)


arximboldi:
Cool, thanks!

PR #15, opened 2019-10-03
Add Todo list examples

arximboldi:


lager/reader.hpp
```diff
@@ -42,6 +42,12 @@ struct reader_mixin
         return attred(member, *this);
     }
 
+    template <typename Xform>
+    auto xf(Xform&& xf)
+    {
+        return xform(xf)(*this);
```
CJBussey: It seems you're missing a forward here with `xf` (also in the writer xf impl)

lager/extra/qt.hpp
```diff
@@ -0,0 +1,85 @@
...
+    type LAGER_QT(name);                                                       \
+    Q_PROPERTY(type name READ name CONSTANT)                                   \
+    const type& name() const { return LAGER_QT(name); }                        \
+    /**/
+
+#define LAGER_QT_SIGNAL_DETAIL(type, name)                                     \
+    Q_SIGNAL void name##Changed(const type& name);                             \
+    ::lager::detail::qt_helper funq__##name##__initHelper__ = [this] {         \
+        ::lager::watch(LAGER_QT(name),                                         \
+                       [this](const type& old, const type& curr) {             \
```
CJBussey: `old` will produce a warning an unused variable here

lager/extra/qt.hpp
```diff
@@ -0,0 +1,85 @@
...
+                       });                                                     \
+        return ::lager::detail::qt_helper{};                                   \
+    }();                                                                       \
+    /**/
+
+/*!
+ * Use in a QObject class declaration to define a property supported
+ * by a lager::reader<> value.
+ */
+#define LAGER_QT_READER(type, name)                                            \
```
CJBussey: Very nice :)

example/todo/qml/main.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+public:
+    Todo(lager::cursor<todo> data)
+        : LAGER_QT(done){data[&todo::done]}
+        , LAGER_QT(text){data[&todo::text].xf(
+              zug::map([](auto&& x) { return QString::fromStdString(x); }),
+              zug::map([](auto&& x) { return x.toStdString(); }))}
+    {}
+
+    LAGER_QT_CURSOR(bool, done);
```
CJBussey: This is very nice :)

lager/extra/qt.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+namespace detail {
+
+struct qt_helper
+{};
+
+} // namespace detail
+
+/*!
+ * Use in a class declaration before any FunQ definition.
```
CJBussey: Is this still true? I can't see you using it anywhere!

arximboldi: Nope :p Good catch!

example/todo/qml/main.cpp
arximboldi: It's ugly to have a macro, but Qt is so verbose...

lager/extra/qt.hpp
arximboldi: Not sure I care :p

lager/reader.hpp
arximboldi: Good catch!

lager/extra/qt.hpp
CJBussey: 😱

example/todo/todo.hpp
```diff
@@ -0,0 +1,39 @@
...
+//
+// This file is part of lager.
+//
+// lager is free software: you can redistribute it and/or modify
+// it under the terms of the MIT License, as detailed in the LICENSE
+// file located at the root of this source code distribution,
+// or here: <https://github.com/arximboldi/lager/blob/master/LICENSE>
+//
+
+#include <immer/flex_vector.hpp>
```
CJBussey: This can be removed, and you should add `<string>` and `<vector>`

lager/extra/qt.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+#define LAGER_QT_SIGNAL_DETAIL(type, name)                                     \
+    Q_SIGNAL void name##Changed(const type& name);                             \
+    ::lager::detail::qt_helper funq__##name##__initHelper__ = [this] {         \
+        ::lager::watch(LAGER_QT(name),                                         \
+                       [this](const type& old, const type& curr) {             \
+                           this->name##Changed(curr);                          \
+                       });                                                     \
+        return ::lager::detail::qt_helper{};                                   \
+    }();                                                                       \
```
CJBussey: What do you think about removing the `;` here so the client code is forced to add it and using `LAGER_QT_READER` and `LAGER_QT_CURSOR`?

arximboldi: I thought about that, I am not sure if I prefer ; or not. What do you prefer?

example/todo/todo.hpp
arximboldi: Good catch :) I actually wanna use `flex_vector`, but need to tweak the `[]` operator of the cursors first.

lager/extra/qt.hpp
CJBussey: I did like the idea to keep the syntax as close to a declaration as possible, however, Qt doesn't enforce it with `Q_PROPERTY`, so using `LAGER_QT_READER` and `Q_PROPERTY` side-by-side might feel a bit annoying (having to remember which needs the `;`). But in that case, I would remove the `;` in your usage example.

arximboldi:
The cursor based Todo QML example is reasonably feature complete. I would like to also add some examples based on Redux (one in QML, another in ImGui), but I will do that in later branches. My next goal (in a sepparate branch) is to add some minimal integration between the Redux and cursor parts, basically making the `store` a `reader` cursor, and using that to replace `current()` and the `view()` functions.

PR #19, opened 2019-10-20
qapplication: add event loop implementation

CJBussey:
I implemented this while creating a QML example using the `lager::store`. I'll commit the example in a separate PR.

### DIFF
--- a/lager/event_loop/qapplication.hpp
+++ b/lager/event_loop/qapplication.hpp
@@ -1,0 +1,48 @@
+//
+// lager - library for functional interactive c++ programs
+// Copyright (C) 2019 Carl Bussey
+//
+// This file is part of lager.
+//
+// lager is free software: you can redistribute it and/or modify
+// it under the terms of the MIT License, as detailed in the LICENSE
+// file located at the root of this source code distribution,
+// or here: <https://github.com/arximboldi/lager/blob/master/LICENSE>
+//
+
+#pragma once
+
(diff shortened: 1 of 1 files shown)


lager/event_loop/qapplication.hpp
```diff
@@ -0,0 +1,48 @@
...
+namespace lager {
+
+struct with_qapplication_event_loop
+{
+    std::reference_wrapper<QApplication> app;
+
+    template <typename Fn>
+    void async(Fn&& fn)
+    {
+        QtConcurrent::run(std::forward<Fn>(fn));
```
arximboldi: It would be nice to have something to garantee that async operations are finished at some point before the desutrction of the store. We did something like that for `iostreams`. We do not need to do it now though.

lager/event_loop/qapplication.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include <QtConcurrentRun>
+
+#include <functional>
+#include <utility>
+
+namespace lager {
+
+struct with_qapplication_event_loop
+{
+    std::reference_wrapper<QApplication> app;
```
arximboldi: I think we could depend in `QCoreApplication` instead, to make it compatible with apps that do not use `QWidgets`, like some QML stuff.

CJBussey: Good point. Fixed.

PR #20, opened 2019-10-20
example: add snake example

CJBussey:
- demonstrate how to use the lager::store with qt/qml

### DIFF
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -92,12 +92,15 @@ endif()
 # ===========
 
 find_package(Qt5Core)
+find_package(Qt5Concurrent)
 find_package(Qt5Qml)
 find_package(Qt5Gui)
 find_package(Qt5Widgets)
 find_package(Qt5QuickControls2)
 
-if (Qt5Core_FOUND AND Qt5Gui_FOUND AND Qt5Widgets_FOUND AND Qt5Qml_FOUND AND Qt5QuickControls2_FOUND)
+find_package(Boost 1.56)
+
+if (Qt5Core_FOUND AND Qt5Concurrent_FOUND AND Qt5Gui_FOUND AND Qt5Widgets_FOUND AND Qt5Qml_FOUND AND Qt5QuickControls2_FOUND AND Boost_FOUND)
   add_executable(todo-qml EXCLUDE_FROM_ALL
     todo/qml/main.cpp)
   set_target_properties(todo-qml PROPERTIES AUTOMOC YES)
@@ -105,6 +108,15 @@ if (Qt5Core_FOUND AND Qt5Gui_FOUND AND Qt5Widgets_FOUND AND Qt5Qml_FOUND AND Qt5
   target_compile_definitions(todo-qml PRIVATE
     LAGER_TODO_QML_DIR="${CMAKE_SOURCE_DIR}/example/todo/qml")
   add_dependencies(examples todo-qml)
+
+  add_executable(snake-qml EXCLUDE_FROM_ALL 
+    snake/model.cpp snake/qml/main.cpp snake/qml/qmodel.cpp)
+  set_target_properties(snake-qml PROPERTIES AUTOMOC YES)
+  target_link_libraries(snake-qml lager Qt5::Core Qt5::Concurrent Qt5::Gui Qt5::Widgets Qt5::Qml Qt5::QuickControls2)
+  target_include_directories(snake-qml PRIVATE ${Boost_INCLUDE_DIRS})
+  target_compile_definitions(snake-qml PRIVATE
+    LAGER_SNAKE_QML_DIR="${CMAKE_SOURCE_DIR}/example/snake/qml")
+  add_dependencies(examples snake-qml)
 else()
   message(STATUS "Disabling Qt based examples")
 endif()
--- a/example/snake/model.hpp
+++ b/example/snake/model.hpp
@@ -1,0 +1,87 @@
+//
+// lager - library for functional interactive c++ programs
+// Copyright (C) 2019 Carl Bussey
+//
+// This file is part of lager.
+//
+// lager is free software: you can redistribute it and/or modify
+// it under the terms of the MIT License, as detailed in the LICENSE
+// file located at the root of this source code distribution,
+// or here: <https://github.com/arximboldi/lager/blob/master/LICENSE>
+//
+
+#pragma once
+
+#include <boost/variant/variant.hpp>
+
+#include <cstdlib>
+#include <random>
+#include <utility>
+#include <vector>
+
+namespace sn {
+
+struct go_left
+{};
+struct go_right
+{};
+struct go_up
+{};
+struct go_down
+{};
+struct reset
+{};
+struct tick
+{};
+using action_t = boost::variant<go_left, go_right, go_up, go_down, reset, tick>;
+
+using point_t = std::pair<int, int>;
+
+inline int x(point_t p) { return p.first; }
+
+inline int y(point_t p) { return p.second; }
+
+enum class direction
+{
+    left,
+    up,
+    right,
+    down
+};
+
+struct snake_model
+{
+    using body_t = std::vector<point_t>;
+    body_t body{};
+    direction dir{};
+};
+
+struct game_model
+{
+    snake_model snake;
+    point_t apple_pos{};
+    bool over;
+
+    static constexpr int width{25};
+    static constexpr int height{25};
+};
+
+struct app_model
+{
+    std::mt19937 gen;
+    std::uniform_int_distribution<> dist;
+
+    game_model game;
+};
+
+template <typename Func>
+inline point_t random_apple_pos(Func&& random)
+{
+    return {random(), random()};
+}
+
+app_model make_initial(int seed);
+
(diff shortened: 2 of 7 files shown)


example/snake/actions.hpp
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+#include <boost/variant/variant.hpp>
```
CJBussey: Neote: I'm using boost variant over std variant, as std variant depends on the macOS 10.14 runtime (for the exception definition used in std::visit). I wrote this on macOS 10.13.

example/snake/actions.hpp
```diff
@@ -0,0 +1,53 @@
...
+struct go_down
+{};
+struct reset
+{};
+using view_action_t = boost::variant<go_left, go_right, go_up, go_down, reset>;
+
+// all actions
+struct tick
+{};
+using action_t = detail::append_to_variant_t<view_action_t, tick>;
```
arximboldi: Interesting, normally I use subvariants instead.

example/snake/main.cpp
```diff
@@ -0,0 +1,81 @@
...
+    auto store =
+        lager::make_store<action_t>(std::move(initial_state),
+                                    reduce,
+                                    lager::with_qapplication_event_loop{app});
+
+    auto* timer = new QTimer(&app);
+    QObject::connect(timer,
+                     &QTimer::timeout,
+                     [c = lager::context<tick>{store}] { c.dispatch(tick{}); });
+    timer->start(initial_update_period);
```
arximboldi: Not sure I like all this `QTimer` based ticking. Another way is to call `tick()` from QML. In QML you can do per-frame ticks using an animation. At least I would try to abstract it away so it does not pollute main in a generic function (something like a `exec_with_tick([] (auto time_delta){ ... })`)

example/snake/main.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    QGame game{store};
+    watch(store, [=](auto&& old, auto&& state) {
+        if (!old.over && state.over) {
+            timer->stop();
+            return;
+        }
+
+        auto oldSpeed = int((old.snake.body.size() - initial_snake_size) / 5);
```
arximboldi: - `old_snake`?
- C-style casts vs `static_cast`?

example/snake/qmodel.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+#include <lager/context.hpp>
+
+#include <QAbstractListModel>
+#include <QObject>
+#include <QPoint>
+
+namespace sn {
+
+class QSnakeBody : public QAbstractListModel
```
arximboldi: Do not prefix with `Q` our own types.

example/snake/reduce.hpp
```diff
@@ -0,0 +1,143 @@
...
+
+                if (std::any_of(body.begin() + 1,
+                                body.end(),
+                                [&](const auto& p) { return p == head; }) ||
+                    !detail::in_bounds(head)) {
+                    game.over = true;
+                    return game;
+                }
+
+                if (prevBack == game.apple_pos) {
```
arximboldi: `prevBack` vs `prev_back`? ;)

example/snake/main.cpp
```diff
@@ -0,0 +1,81 @@
...
+        auto speed    = int((state.snake.body.size() - initial_snake_size) / 5);
+
+        if (oldSpeed != speed || !timer->isActive()) {
+            timer->start(std::max(initial_update_period -
+                                      update_period_decrement * speed,
+                                  min_update_period));
+        }
+    });
+    watch(store, [&](auto&& /*old*/, auto&& state) {
+        game.update(std::forward<decltype(state)>(state));
```
arximboldi: Interesting that you are calling update to pass the value instead of passing a cursor. I guess this is part of the exercise?

example/snake/main.cpp
```diff
@@ -0,0 +1,81 @@
...
+    });
+    game.update(store.get());
+
+    auto* qmlContext = engine.rootContext();
+    qmlContext->setContextProperty(QStringLiteral("game"), &game);
```
arximboldi: I do not like using `setContextProperty`. I prefer to register a type that the user instantiates, this root type can hold the `store` and do part of the setup that is done here. Helps keeping `main.cpp` and makes the "model" easy to extract into a QML plugin or library. See the `todo-qml` for an example.

example/snake/model.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+enum class direction
+{
+    left,
+    up,
+    right,
+    down
+};
+
+struct snake_t
```
arximboldi: Not fan of suffixing all types with `_t`, I only use it for aliases.

example/snake/model.hpp
```diff
@@ -0,0 +1,73 @@
...
+    bool over;
+
+    static constexpr int width{25};
+    static constexpr int height{25};
+};
+
+inline point_t random_apple_pos()
+{
+    static const auto seeder = [] {
+        std::srand(std::time(nullptr));
```
arximboldi: :warning: `std::time` and `std::srand` are not pure functions! Do not use in the reducer!

You have two options:
1. use an effect to compute the random number, then return it back into the game with an action, or 
2. put the pseudo-random sequence as part of the model, for example using the stuff from `<random>`. You can get a entrophised seed before creating the store.

example/snake/model.hpp
```diff
@@ -0,0 +1,73 @@
...
+        return true;
+    }();
+    std::ignore = seeder;
+    return {float(std::rand()) / RAND_MAX * game_t::width,
+            float(std::rand()) / RAND_MAX * game_t::height};
```
arximboldi: `static_cast` my friend.

example/snake/qmodel.cpp
```diff
@@ -0,0 +1,79 @@
...
+        return as_qpoint(data_[index.row()]);
+    return {};
+}
+
+QGame::QGame(context_t context)
+    : context_{std::move(context)}
+    , snakeBody_{new QSnakeBody(this)}
+{}
+
+void QGame::update(game_t game)
```
arximboldi: If we call the other function `update()` this could be called `view`. You could consider having all these qt based models get a reader cursor and do the watching themselves. Would simplify some of this code for sure.

example/snake/main.cpp
CJBussey: Personally, I find `static_cast` way too verbose for casting between fundamentals. Note this isn't a C-style cast, but a functional-style cast. Essentially does the same, though. If you have a strong opinion, I can change it.

example/snake/reduce.hpp
CJBussey: I like switching it up every now and then... keeps the reader on their toes!

example/snake/main.cpp
CJBussey: Yes, that's correct. As mentioned in our chat, this was an experiment to play with the store without the cursors and with keeping the Qt/QML side to a minimum, essentially only using it for GUI interactions. I'll look into another example with the cursors, though.

example/snake/model.hpp
CJBussey: Actually just did this because gcc complained about using the same name for the variable and type. I've now renamed it to something more verbose, but I prefer to be less verbose but still understandable for the variable (i.e., `game` instead of `g`).

example/snake/model.hpp
CJBussey: Thanks for pointing this out. In case anybody else is reading this, I thought it would be worth adding your elaboration here (hope you don't see this as a betrayal of your trust by sharing a private conversation!):

> Testability is one reason.
> But yeah, the idea is that when you replay the same sequence of actions you get the same result
> for example, the time travelling debugger now stores previous states
> but in theory it should not need to
> it can just save snapshots from time to time and interpolate replaying the actions in between

example/snake/qmodel.cpp
CJBussey: I've renamed the other `update`. `view` felt a little bit strange to me as a member function. From the object oriented perspective, it feels a little bit confusing, as it doesn't describe how I'm interacting with the object. (From a functional perspective, it makes sense.)

As mentioned in a previous comment - I'll definitely have a play around more with the cursors, but I wanted to keep them out of this example.

example/snake/main.cpp
CJBussey: I moved it to a QML timer - much simpler.

example/CMakeLists.txt
```diff
@@ -92,19 +92,31 @@ endif()
...
 
 find_package(Qt5Core)
+find_package(Qt5Concurrent)
 find_package(Qt5Qml)
 find_package(Qt5Gui)
 find_package(Qt5Widgets)
 find_package(Qt5QuickControls2)
 
-if (Qt5Core_FOUND AND Qt5Gui_FOUND AND Qt5Widgets_FOUND AND Qt5Qml_FOUND AND Qt5QuickControls2_FOUND)
+find_package(Boost 1.56)
```
arximboldi: We already do a `find_boost` in the main CMakeLists.txt I think.

arximboldi:
Fixed a couple of details... it is working now, super nice!!!

PR #21, opened 2019-10-22
Bugfix/example dont reseed generator

CJBussey:
While playing around with the Snake game, I noticed that the randomness of the apple position on subsequent restarts of the game was off. I noticed this was down to the funky seeding I was doing when restarting. Now when `reset` is called, we only update the game state.


arximboldi:
Good catch! :+1:

PR #23, opened 2019-10-29
missing include to <memory>

XzoRit:
This lib did not compile because of a missing include to memory.


arximboldi:
Cool, thanks for the fixes!

PR #24, opened 2019-11-02
Fix typos in online docs (model.rst)

jimbo1qaz:
Also it's usually called "database" not "data-base". Should I change that too?

I've seen "datastore" and "data store" being used online. Firefox thinks "datastore" is a typo.


arximboldi:
Cool! Thanks for the fixes. Yeah, I'd be happy to take a change for data-base -> database.

PR #27, opened 2019-11-17
Make lager core relocatable

ltjax:
Coding LAGER_PREFIX_PATH into the sources prevents the code from being properly relocatable. It seems to be used only for `lager::resources_path()` and its copy pasta in the debugger. As far as I can tell, the function is only used the examples.
For the debugger, the final application should be able to say where the resources are anyways, so I propose to inject the path via the constructor.
For the examples, the current method is okay, so I just moved the code depending on `LAGER_PREFIX_PATH` there, and injected it via `target_compile_definitions()` instead of a `configure_file()` step. 

**Warning:** I did not compile the examples yet, because I'm missing libhttpserver on my mac. I will try that later, but I guess the CI will also test that.


arximboldi:
I like the goal of removing the configure step to make it easier to install the library by just copying the sources. I do not like the newly added `examples_common` namespace, I would prefer the examples to be more isolated from eachother.

I think we can get away with making the `http_debug_server` still trying to see if `ifdef LAGER_PREFIX_PATH`, and if the environment variable `LAGER_RESOURCS_PATH` is defined. We already define the environment variable in the `nix-shell` anyways.

arximboldi:
Now, that would be, instead of having the `config.hpp`, the `http_debug_server` checks the `ifdef`, which may be optionally passed as a `-D` flag at compile time or not.

ltjax:
Alright, I can see that isolated examples are a good thing. But the path is also used for a few font files outside of the debugger. In that case, maybe it's a good idea to have copies of the path selection code in each sample that uses it and get rid of example-common that way. Bot much value in DRY for that function anyways.
That way, the examples are even more obvious on what paths it looks for, since that is usually a thing that breaks easily on someone else's machine.

ltjax:
How's that?

arximboldi:
I would prefer if the `http_debug_server` applied that logic itself to find a "default" resources location (the argument could still be there, optional), so only examples that use their own resources (it seems there is only one, using a font) really need to play with defining `resources_path()` and `LAGER_PREFIX_PATH`. If you are worried about people forgetting to define that, you can throw an exception if `LAGER_PREFIX_PATH` is not defined and `LAGER_RESOURCES_PATH` is not in the environment and now path is passed explicitly.

This would allow it to be a non breaking change. Also, I was just considering (I am no CMake expert) is it possible to make the Cmake target that is installed to include public defines?

The idea is that if I install Lager with `make install`, then when I do:
```
find_package(lager)
add_executable(foo ...)
target_link_libraries(foo lager)
```
Then I get `-DLAGER_PREFIX_PATH="..."` set for `foo`.

This would turn this into a fully non breaking API change.

ltjax:
I really don't want a #define that has a side effect on code in a header. That is the very definition of evil preprocessing, and also the reason why C++20 modules do not support that. Very easy to violate ODR with that.
Also, the path that you build from the LAGER_PREFIX_PATH is very Linux-y. Configuring resources paths is absolutely an application concern that a library should not dictate. If you want to be really clean here, you should even inject the whole process of opening a file, because the data might end up in a .zip or something for the final application.
TBH, I think using constructor injection solves both problems neatly and is an improvement here, and well worth an API break at this point. Potentially supporting modules alone in the future makes it worth it.

ltjax:
Oh, and sure, you can transfer defines via CMake, but that'd defeat the purpose of making the core library relocatable. You're just storing the #define somewhere else then.

ltjax:
If you really really really really want the old ctor-API to work, I guess we could inject the LAGER_PREFIX_PATH at the point of instantiation via a default argument, but people will get weird error messages.

arximboldi:
Hmmm. I do understand your arguments, the thing is, I want the easy use-case to be easy. Note these are not resources from the user, these are resources from the library.

So if I install the library with `make install`, why do I need to tell the library again where its resources live?

So my goal is: if just I install the library with `make install` and then find it, instantiating the `http_debug_server` "just works". If I install the library by other mechanism, then I can just tell it where the resources are with the extra optional argument, or with the environment variable. Do you think there is a way to achieve this behavior that satisfies you?

ltjax:
You're welcome. Go ahead, I think this is done.

PR #36, opened 2020-03-04
Extend lenses/cursors API

Siapran:
This branch does a few things:
- misc fixes to get the examples from [Squaring the Circle](https://sinusoid.es/talks/meetingcpp19) running on MSVC 2017
- introduce a type erased lens
- enable the `cursor[lens]` syntax
- implement unfocused lenses (lenses that return optionals instead of default constructed values when the target value is not present)
- partial implementation for automatically lifting smart lenses passed into `cursor<optional<T>>[]`

So far, the only tests built, ran and passed on MSVC are `lenses.cpp`, `state.cpp`, `cursor.cpp`, and `any_lens.cpp`.
`xformed.cpp` breaks really bad on MSVC right now.
The purpose of this PR is mostly to review the work done so far.


lager/lenses.hpp
```diff
@@ -123,51 +128,103 @@ auto attr(Member member)
...
+            using Part = std::optional<std::decay_t<decltype(whole.at(key))>>;
+
+            return f([&]() -> Part {
                 try {
-                    return std::forward<decltype(p)>(p).at(key);
-                } catch (std::out_of_range const&) {
-                    return std::decay_t<decltype(p.at(key))>{};
+                    return std::forward<decltype(whole)>(whole).at(key);
+                } catch (std::out_of_range const&) { return std::nullopt; }
+            }())([&](Part part) {
```
Siapran: A copy is the easy way out here. I tried really hard to avoid constructing optionals when passed a value but it ended up being way [overkill](https://gist.github.com/Siapran/6e6a87ac5043e13a89bbb289dad4aacf)...

lager/lenses.hpp
```diff
@@ -123,51 +128,103 @@ auto attr(Member member)
...
 
+/*!
+ * `X -> Lens<[X], X>`
+ */
+template <typename T>
+auto fallback(T&& t) {
+    return zug::comp([t = std::forward<T>(t)](auto&& f) {
+        return [&, f](auto&& whole) {
+            return f(LAGER_FWD(whole).value_or(std::move(t)))(
+                [&](auto&& x) { return LAGER_FWD(x); });
```
Siapran: see [this comment](https://gist.github.com/Siapran/6e6a87ac5043e13a89bbb289dad4aacf#file-overkill-hpp-L116)

lager/lenses.hpp
```diff
@@ -123,51 +128,103 @@ auto attr(Member member)
...
+                            return LAGER_FWD(whole);
+                        }
+                    });
+            } else {
+                return f(Part{std::nullopt})(
+                    [&](auto&&) { return LAGER_FWD(whole); });
+            }
+        };
+    });
+}
```
Siapran: This needs a cleaner implementation.

Siapran:
The `xformed.cpp` test fails to compile, I didn't get to adapt it for the new versions of `at`.

lager/lenses.hpp
```diff
@@ -123,51 +128,103 @@ auto attr(Member member)
     });
 }
 
+/*!
+ * `Key -> Lens<{X}, [X]>`
+ */
```
Siapran: The `at(key)` family of lenses now returns an optional instead of a default constructed value if the target is not present.

Siapran:
working on fixing the unit tests on Linux

Siapran:
Updated example and unit test so they build.
More work needed on `xformed.cpp`, one of the tests fails.

Siapran:
Aaaaand done!

lager/detail/smart_lens.hpp
```diff
@@ -46,9 +58,32 @@ struct smart_lens_base
...
+using viewed_t = decltype(
+    std::declval<Lens>()(functor_wrapper_t{})(std::declval<T>()).value);
+
+template <typename T>
+struct smart_lens<std::optional<T>>
+{
+    template <typename U>
+    static auto make(U &&u) {
+        // don't lift lenses that can already handle optionals
+        if constexpr (zug::meta::is_detected<viewed_t, U, std::optional<T>>::value) {
```
Siapran: This doesn't work quite right yet.
It detects if the given type is a lens, but not if it can handle optionals specifically...

Siapran: yeah I don't think I can SFINAE viewing the lens right because [return type deduction doesn't work with SFINAE](https://en.cppreference.com/w/cpp/language/function#Return_type_deduction) and pretty much all the lenses use return type deduction _everywhere_...

I don't know how to make this work right. I might have to just accept that `lens` in `cursor[lens]` has to handle optionals explicitly.

lager/lenses.hpp
```diff
@@ -123,51 +128,116 @@ auto attr(Member member)
...
+
+/*!
+ * `(Lens<W, P> | Lens<W, [P]>) -> Lens<[W], [P]>`
+ */
+template <typename Lens>
+auto optlift(Lens&& lens) {
+    return zug::comp([lens = std::forward<Lens>(lens)](auto&& f) {
+        return [&, f](auto&& whole) {
+            using Whole = std::decay_t<decltype(whole)>;
+            using Part  = std::optional<std::decay_t<decltype(::lager::view(
```
Siapran: Oops, this actually only does `Lens<W, P> -> Lens <[W], [P]>`.
Should I make two separate functions for `Lens<W, [P]> -> Lens <[W], [P]>` and `Lens<W, P> -> Lens <[W], [P]>` or just keep them merged as one function.
Or even provide `optlift` as a convenience for both, and add two others for explicitly handling one or the other?

Siapran: Went with all three, although I'm weary of `optlift` being too much of a footgun.

lager/lenses.hpp
```diff
@@ -237,6 +237,77 @@ auto optlift(Lens&& lens) {
...
+                        }
+                    });
+            } else {
+                return f(Part{std::nullopt})(
+                    [&](auto&&) { return LAGER_FWD(whole); });
+            }
+        };
+    });
+}
+
```
Siapran: Could definitely use some factoring.

Siapran:
Apparently, `any_lens` doesn't work right if you construct it from an lvalue reference to another lens... (works fine with rvalues though)

Siapran:
Fixed in fa15143.
Need to write unit tests for this.

lager/lenses.hpp
Siapran: Done at 98cc141.

Siapran:
I'm on the edge for using a lens-specific type for signaling lack of focus.
Something like `lager::lens::focus<T>`  and `lager::lens::unfocused` instead of `std::optional` would avoid awkwardness when using models that have optionals as legitimate values.
On the other hand, being explicit about what kind of lens manipulation you're doing can't hurt, and I can already see people storing `focus<T>` in their model and blowing their feet clean off with the footgun that is `optlift`.

So yeah, either use `focus<T>` or remove `optlift` in favor of `optmap` and `optbind`.

lager/lenses.hpp
Siapran: Still not sure if this is the best way.

Siapran:
looking into reimplementing `opt*(lens)` in terms of functors instead of manually lifting them like I'm doing right now.

lager/detail/smart_lens.hpp
```diff
@@ -27,9 +29,19 @@ class flex_vector;
...
 namespace detail {
 
+template <typename T, typename Key>
+using at_t = std::decay_t<decltype(std::declval<T>().at(std::declval<Key>()))>;
+
+template <typename T>
 struct smart_lens_base
 {
-    template <typename Key>
+    template <typename Lens, std::enable_if_t<!zug::meta::is_detected<at_t, T, Lens>::value, int> = 0>
```
arximboldi: I would prefer to detect this case by using a `is_detected` expression that explicitly checks if `Lens` supports `set(l, declval<T>(), view(l, declval<T>()))`.

lager/detail/smart_lens.hpp
```diff
@@ -46,9 +58,32 @@ struct smart_lens_base
...
 {};
 
+struct functor_wrapper_t {
+    template <typename T>
+    const_functor<T> operator()(T&& t) const noexcept;
+};
+
+template <typename Lens, typename T>
+using viewed_t = decltype(
+    std::declval<Lens>()(functor_wrapper_t{})(std::declval<T>()).value);
```
arximboldi: Why not: `decltype(view(declval<Lens>(), declval<T>))`?

lager/detail/smart_lens.hpp
arximboldi: AFAIK, SFINAE works with type deduction if you use `decltype(auto)` I believe, maybe we just need to tweak the lenses to use that in the right places.

lager/lenses.hpp
```diff
@@ -109,11 +112,14 @@ auto getset(Getter&& getter, Setter&& setter)
...
 
+/*!
+ * `(Part Whole::*) -> Lens<Whole, Part>`
+ */
 template <typename Member>
 auto attr(Member member)
 {
-    return zug::comp([=](auto&& f) {
-        return [&, f](auto&& p) {
+    return zug::comp([member](auto&& f) {
```
arximboldi: Why change to explicit here?

lager/lenses.hpp
```diff
@@ -123,51 +129,169 @@ auto attr(Member member)
...
+/*!
+ * `Key -> Lens<{X}, [X]>`
+ */
 template <typename Key>
-auto at(Key key)
-{
-    return zug::comp([=](auto&& f) {
-        return [f, &key](auto&& p) {
-            return f([&] {
+auto at(Key key) {
```
arximboldi: I like this new version, but I think the old behavior is more consistent with what C++ often does, for example with `operator[]`. I would add a new version called `at_opt` for this.

lager/lenses.hpp
```diff
@@ -123,51 +129,169 @@ auto attr(Member member)
...
             });
         };
     });
 }
 
+/*!
+ * `Key -> Lens<{X}, [X]>`
+ */
 template <typename Key>
-auto at_i(Key key)
```
arximboldi: Same.

lager/lenses.hpp
```diff
@@ -123,51 +129,169 @@ auto attr(Member member)
...
+            });
+        };
+    });
+}
+
+/*!
+ * `X -> Lens<[X], X>`
+ */
+template <typename T>
+auto fallback(T&& t) {
```
arximboldi: Rename to `value_or`.

lager/lenses.hpp
```diff
@@ -123,51 +129,169 @@ auto attr(Member member)
...
+            return f(LAGER_FWD(whole).value_or(std::move(t)))(
+                [&](auto&& x) { return LAGER_FWD(x); });
+        };
+    });
+}
+
+/*!
+ * `() -> Lens<[X], X>`
+ */
+auto fallback() {
```
arximboldi: Rename to `value_or_` and flatten one level of functions here (see `ZUG_INLINE_CONSTEXPR`).

lager/lenses.hpp
```diff
@@ -123,51 +129,169 @@ auto attr(Member member)
...
+ * `(Lens<W, P> | Lens<W, [P]>) -> Lens<[W], [P]>`
+ */
+template <typename Lens>
+auto optlift(Lens&& lens) {
+    return zug::comp([lens = std::forward<Lens>(lens)](auto&& f) {
+        return [&, f = LAGER_FWD(f)](auto&& whole) {
+            return detail::opt_impl<detail::to_opt>(LAGER_FWD(whole), lens, f);
+        };
+    });
+}
```
arximboldi: Not sure about the naming of all above. I may prefer `map_opt`, `bind_opt`, `lift_opt`. Also, it sounds like `map_opt` and `bind_opt` are redundant, since `lift_opt` does all. If that is the only version, maybe `with_opt` is better (a bit less functional sounding, but...)

arximboldi: Also if you only leave one version you can inline `opt_impl` making it all a bit cleaner.

lager/lenses.hpp
```diff
@@ -123,51 +129,169 @@ auto attr(Member member)
...
+/*!
+ * `Key -> Lens<{X}, [X]>`
+ */
 template <typename Key>
-auto at_i(Key key)
-{
-    return zug::comp([=](auto&& f) {
-        return [f, &key](auto&& p) {
-            return f([&] {
+auto at_i(Key key) {
```
arximboldi: Also... I have been doubting a lot about this function, maybe it should be `at_immer`.

lager/lenses.hpp
```diff
@@ -123,51 +129,169 @@ auto attr(Member member)
...
+auto optlift(Lens&& lens) {
+    return zug::comp([lens = std::forward<Lens>(lens)](auto&& f) {
+        return [&, f = LAGER_FWD(f)](auto&& whole) {
+            return detail::opt_impl<detail::to_opt>(LAGER_FWD(whole), lens, f);
+        };
+    });
+}
+
+
 } // namespace lens
```
arximboldi: This file is growing... Maybe we should split it into multiple files in `lager/lens/...`.

test/cursor.cpp
```diff
@@ -108,3 +111,32 @@ TEST_CASE("values, scoped watching")
...
+using boost::fusion::operators::operator==;
+using boost::fusion::operators::operator!=;
+
+TEST_CASE("zooming, unfocused")
+{
+    state<std::vector<person>> st{{person{{5, 4}, "juanpe"}}};
+    auto p  = st[0];
+    auto n1 = p[&person::name];
+//    auto n2 = p[lens::attr(&person::name)];
+//    reader<std::optional<std::string>> n2 = p[lens::attr(&person::name)];
```
arximboldi: What is this commented stuff about?

lager/any_lens.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+    Whole set(Whole const& w, Part const& p) const override {
+        return ::lager::set(value, w, p);
+    }
+};
+
+} // namespace detail
+
+template <typename Whole, typename Part>
+class any_lens : zug::detail::pipeable {
```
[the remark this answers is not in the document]
Siapran: agreed with `lens<>` instead of `any_lens<>`, the main motivation was to discourage its use over non type erased alternatives.
Not sure about the name collision problem, I'm kind of concerned about introducing too many breaking changes.

lager/any_lens.hpp
```diff
@@ -0,0 +1,66 @@
...
+              typename std::enable_if<
+                  !std::is_same_v<std::decay_t<Lens>, std::decay_t<any_lens>>,
+                  int>::type = 0>
+    any_lens(Lens&& lens)
+        : holder_{new detail::lens_holder<std::decay_t<Lens>, Whole, Part>{
+              std::forward<Lens>(lens)}} {}
+
+    template <typename F>
+    auto operator()(F &&f) const {
+        return [&, f = std::forward<F>(f)](auto&& p) {
```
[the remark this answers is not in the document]
Siapran: agreed.

lager/commit.hpp
```diff
@@ -17,6 +17,21 @@
 
 namespace lager {
 
+namespace detail {
+
+template <typename RootCursorT>
+void send_down_root(RootCursorT&& root)
+{
+    noop((detail::access::roots(std::forward<RootCursorT>(root))->send_down(), 0));
```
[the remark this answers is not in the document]
Siapran: if I recall correctly, removing the `noop` broke this for some reason. I'll have to give it another try.

lager/detail/smart_lens.hpp
Siapran: so the problem here is that `set`, `view`, all the lenses, and just about everything we're using are all using return type deduction, [which unfortunately _doesn't_ work for SFINAE]...(https://en.cppreference.com/w/cpp/language/function#Return_type_deduction)
for that reason, I decided to switch on the `at_t` instead.

lager/detail/smart_lens.hpp
Siapran: same as above, can't sfinae on lenses...

lager/detail/smart_lens.hpp
Siapran: if we do decide to make all our API SFINAE friendly to enable more metaprogramming stuff, it means we need to enforce explicit return types on lambdas passed to things like `lens::getset`.

lager/lenses.hpp
Siapran: no real reason beyond style.

lager/lenses.hpp
Siapran: I disagree with this on several points:
- [`vector::operator[]`](https://en.cppreference.com/w/cpp/container/vector/operator_at) does _not_ return a default constructed element when the index is out of bounds. accessing an OOB index is, in fact undefined behavior.
- [`map::operator[]`](https://en.cppreference.com/w/cpp/container/map/operator_at) _does_ return a default constructed element, _but also adds it to the collection_. your implementation explicitly forbids writing at an OOB index.
- the name of this lens is `at(idx)`, which, both in the `std` api and any other api, is able to signal the abscence of a value (`std::out_of_bounds`, `Maybe`, etc).

Maybe having a lens named `at_or(idx, [value])` could be the solution here?

lager/lenses.hpp
Siapran: maybe  `lager::lens::immer::at()`?
this way we could also have lenses for other immer containers like `lens::immer::unbox()`

lager/lenses.hpp
Siapran: agreed.

lager/lenses.hpp
Siapran: agreed.

lager/lenses.hpp
Siapran: agreed for the naming changes. `with_opt` also does sound better.
the main problem with `with_opt` is that it may yield incorrect semantics if used on a model that contains legitimate optionals. this is the reason I'm also providing `map_opt` and `bind_opt`. it is also the reason I was thinking of deprecating `with_opt` altogether or specifying a different type, `lager::lens::focus<T>` instead of `std::optional<T>` to signal a lens that may be out of focus (aka, absent value).

lager/lenses.hpp
Siapran: agreed.

test/cursor.cpp
Siapran: this is the syntax I was trying to SFINAE on, but couldn't because of the return type deduction shenanigans.
these lines won't compile, which is better than nothing at least.

Siapran:
I was trying to implement more convenience lenses, but after encountering [this mess](https://github.com/Siapran/lager/commit/35cd8574deb14e32c53629975c9e72ea928cf4c3), I decided I wanted a better implementation for optics, instead of using `lens` as a catchall for everyting.
I've been reading on [Profunctor Optics](https://github.com/hablapps/DontFearTheProfunctorOptics), and I feel like this might be possible to implement in C++, which would make this whole concept of unfocused lenses a lot less awkward to deal with.

lager/detail/functors.hpp
```diff
@@ -0,0 +1,46 @@
+#pragma once
```
arximboldi, line 1: Not sure if I agree with introducing this new file, since these functors are very ad-hoc and only used as an implementation detail for the lenses `view`,`over`, etc and this is only included from `lenses.hpp`. On the other hand, it is ok maybe, I don't wanna nitpick too much.

lager/lenses/at.hpp
```diff
@@ -0,0 +1,68 @@
...
+        try {
+            (void) whole.at(std::forward<Key>(key));
+            return std::forward<Whole>(whole).set(
+                std::forward<Key>(key), std::forward<Part>(part).value());
+        } catch (std::out_of_range const&) {}
+    }
+    return std::forward<Whole>(whole);
+}
+
+} // namespace detail
```
arximboldi, line 46: Nice way of doing Immer support!

lager/lenses/variant.hpp
```diff
@@ -0,0 +1,45 @@
...
+                    return LAGER_FWD(p);
+                }
+            });
+        };
+    }
+};
+} // namespace detail
+
+template <typename T>
+auto var_at = detail::var_at_t<T>{};
```
arximboldi, line 42: Not sure about the name `var_at`, maybe `at_alternative` or `with_alternative` would be more consistent with `std`? (even though more wordy)

lager/commit.hpp
```diff
@@ -17,6 +17,21 @@
...
+{
+    (void) detail::access::roots(std::forward<RootCursorT>(root))->send_down();
+}
+
+template <typename RootCursorT>
+void notify_root(RootCursorT&& root)
+{
+    (void) detail::access::roots(std::forward<RootCursorT>(root))->notify();
+}
+} // namespace detail
```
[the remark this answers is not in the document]
Siapran: This is a fix for some weird MSVC issue.

lager/detail/functors.hpp
Siapran: yeah you're right, I'm not sure why I split those from lenses.hpp

PR #37, opened 2020-03-05
Work around VS2019 bug for scope of send down

audiofanatic:
Also includes a fix for a noisy warning seen with VS2019


arximboldi:
LGTM! Thanks a lot! :+1:

PR #59, opened 2020-05-13
Add async() to with_qt_event_loop

jarzec:
- Modify with_qt_event_loop to implement async() with possibility
to specify a QThreadPool, while maintaining backward compatibility
- Add a respective test


lager/event_loop/qt.hpp
```diff
@@ -12,29 +12,54 @@
...
+            return std::visit(lager::visitor{
+                       [](std::reference_wrapper<QObject>) {
+                           return QThreadPool::globalInstance();
+                       },
+                       [](std::reference_wrapper<QThreadPool> thread_pool) {
+                           return &thread_pool.get();
+                       }
+                },
+                obj_ref_variant);
+        }
```
jarzec, line 44: This is not a lambda to make MSVC 2017 happy...

arximboldi:
Hey! Nice! I like the approach.

My only concern is that using a `std::variant` seems unnecessary, since `QThreadPool` already inherits from QObject. You can keeep a `reference_wrapper<QObject>` and simply use `dynamic_cast` inside `async()`. An alternative would be to keep two different reference wrappers, one to a QObject and another one to a QThreadPool.

jarzec:
I wanted to keep one reference wrapper for simplicity of use with construction with a single argument.
I am not an expert, but I was aiming at something closer to compile time resolution. In fact I could add `constexpr` on the `get_thread_pool_ptr()`. But if you think the `std::variant` is overkill I could go with runtime resolution.

Another question is about the `run_test()` function. I forgot earlier, I will add a comment in the code.

test/event_loop/qt.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+    // Process the set_worker_id_action dispatched by the asynchronous worker
+    QCoreApplication::processEvents();
+    // Perform model update
+    QCoreApplication::processEvents();
+
+    CHECK(reader->main_id == main_thread_id);
+    CHECK(reader->worker_id != std::thread::id{});
+    CHECK(reader->worker_id != main_thread_id);
+}
```
jarzec, line 53: I was thinking of adding a third argument to this function - a lambda that would replace the explicit `QCoreApplication::processEvents();` call. With such a solution this function could be moved into _loop.hpp_ and become reusable for testing other event loops that allow for manual stepping.
I have 0 experience with the boost one, but e.g. would something like this be usable there?

jarzec:
I tried out both your suggestions (using `dynamic_cast` or another `std::reference_wrapper`) and decided the latter one was more elegant. Using a separate `std::reference_wrapper<QThreadPool>` (defaulting to the global instance) separates the responsibilities of providing context for `post()` and of providing a thread pool for `async()`.

arximboldi:
I agree that's the cleanest option. Thanks!

test/event_loop/qt.cpp
arximboldi: I think it's ok not to abstract that away. It's nice to keep tests as explicit as possible.

PR #60, opened 2020-05-24
update boost deps

asa:
missing boost intrusive/list.hpp


arximboldi:
Ooops! Thanks a lot @asa!

PR #63, opened 2020-07-14
MSVC nodes fix

Siapran:
This branch addresses two issue encountered on MSVC:
- Some type aliases had some trouble getting through. a workaround was found to fix this.
- A moved from shared pointer was being deferenced because of some reordering done by MSVC. This was fixed by calling the relevant function by reference, and not by value, allowing the `current_from(parents)` in `lens_nodes.hpp` and `merge_nodes.hpp` to be evaluated before the pointer is moved from.


lager/lenses/tuple.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * N -> Lens<(P1, ..., Pn), PN>
+ *
+ * Note: works for pairs, tuples, arrays. don't use on variants.
+ */
+template <size_t N>
+auto element = detail::element_t<N>{};
+inline auto first = element<0>;
+inline auto second = element<1>;
```
arximboldi: third, fourth, fifth...? hehe.

lager/lenses/tuple.hpp
```diff
@@ -1,48 +0,0 @@
...
-template <size_t N>
-auto element = detail::element_t<N>{};
-inline auto first = element<0>;
-inline auto second = element<1>;
-
-//! @}
-
-} // namespace lenses
-} // namespace lager
-
```
arximboldi: Why remove? I like it!

Siapran:
I'll include the tuple stuff in another PR when I'll have more time to properly test it. This tuple.hpp file is also incomplete, my WIP version has lens generators such as zip, explode, and a multi attr. I'll include all of these in their own PR very soon.

lager/lenses/tuple.hpp
Siapran: well, first and second are mostly there because of std::pair

lager/lenses/tuple.hpp
Siapran: I didn't want to include it yet, because I don't have extensive enough tests written for it right now. Don't worry though, you'll get it soon!

arximboldi:
Cool, makes sense, looking forward to that future PR! Merging this now :+1:

PR #64, opened 2020-07-17
Introduce tuple lenses

Siapran:
This branch introduces four new lens generators for working with tuples:
* `zip` aka: `Lens<W1, P1>, ..., Lens<Wn, Pn> -> Lens<(W1, ..., Wn), (P1, ..., Pn)>`
* `fan` aka: `Lens<W, P1>, ..., Lens<W, Pn> -> Lens<W, (P1, ..., Pn)>`
* `attr` with multiple non-overlapping members
* `element<N>`, `first` and `second`, for accessing the Nth element of a tuple-like value.

### DIFF
--- a/lager/lenses/tuple.hpp
+++ b/lager/lenses/tuple.hpp
@@ -1,10 +1,10 @@
 #pragma once
 
-#include <tuple>
-#include <type_traits>
-
 #include <lager/lenses.hpp>
 #include <lager/lenses/attr.hpp>
+
+#include <tuple>
+#include <type_traits>
 
 namespace lager {
 namespace lenses {
--- a/lager/lenses/tuple.hpp
+++ b/lager/lenses/tuple.hpp
@@ -148,7 +148,7 @@ auto zip(Lenses&&... lenses)
  * overwrite sequential writes to the part. just don't do it.
  */
 template <typename... Lenses>
-auto explode(Lenses&&... lenses)
+auto fan(Lenses&&... lenses)
 {
     return zug::comp([lens_tuple =
                           std::make_tuple(LAGER_FWD(lenses)...)](auto&& f) {
@@ -168,13 +168,13 @@ auto explode(Lenses&&... lenses)
 /**
  * (Part Whole::*)... -> Lens<Whole, (Part...)>
  *
- * Note: for the same reason as detailed in explode, the members
+ * Note: for the same reason as detailed in fan, the members
  * should be distinct from each other.
  */
 template <typename... Member>
 auto attr(Member... member)
 {
-    return explode(attr(member)...);
+    return fan(attr(member)...);
 }
 
 /**
--- a/lager/lenses/tuple.hpp
+++ b/lager/lenses/tuple.hpp
@@ -1,0 +1,194 @@
+#pragma once
+
+#include <tuple>
+#include <type_traits>
+
+#include <lager/lenses.hpp>
+#include <lager/lenses/attr.hpp>
+
+namespace lager {
+namespace lenses {
+namespace detail {
+
+template <typename T>
+struct templated_type_contructor;
+
+template <template <typename...> typename T, typename... Params>
+struct templated_type_contructor<T<Params...>>
+{
+    template <typename... Ts>
+    static auto construct(Ts&&... ps)
+    {
+        return T<std::remove_reference_t<Ts>...>{std::forward<Ts>(ps)...};
+    }
+};
+
+template <size_t I, typename... Tuples>
+decltype(auto) multiget(Tuples&&... tuples)
+{
+    return std::forward_as_tuple(std::get<I>(std::forward<Tuples>(tuples))...);
+};
+
+template <typename ResType, typename F, typename... Tuples, size_t... Seq>
+decltype(auto) apply_zip_impl(std::index_sequence<Seq...>, F&& f, Tuples&&... t)
+{
+    return templated_type_contructor<ResType>::construct(std::apply(
+        std::forward<F>(f), multiget<Seq>(std::forward<Tuples>(t)...))...);
+}
+
+template <typename ResType, typename F, typename... Tuple>
+decltype(auto) apply_zip(F&& f, Tuple&&... t)
+{
+    constexpr size_t len =
+        std::min({std::tuple_size_v<std::remove_reference_t<Tuple>>...});
+    return apply_zip_impl<ResType>(std::make_index_sequence<len>{},
+                                   std::forward<F>(f),
+                                   std::forward<Tuple>(t)...);
+}
+
+namespace fold_op {
+template <typename Init, typename Elem>
+decltype(auto) operator+(Init&& init, Elem&& elem)
+{
+    auto&& [lens, part] = LAGER_FWD(elem);
+    return ::lager::set(LAGER_FWD(lens), LAGER_FWD(init), LAGER_FWD(part));
+}
+} // namespace fold_op
+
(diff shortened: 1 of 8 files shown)


lager/lenses/tuple.hpp
```diff
@@ -0,0 +1,194 @@
...
+
+template <template <typename...> typename T, typename... Params>
+struct templated_type_contructor<T<Params...>>
+{
+    template <typename... Ts>
+    static auto construct(Ts&&... ps)
+    {
+        return T<std::remove_reference_t<Ts>...>{std::forward<Ts>(ps)...};
+    }
+};
```
arximboldi, line 24: It seems like the only thing this is really doing is the `remove_reference_t`. This could be done way above the call stack in a less confusing way and get rid of `templated_type_constructor`. Also not sure why you want to `remove_reference` at all...

lager/lenses/tuple.hpp
```diff
@@ -0,0 +1,194 @@
...
+} // namespace fold_op
+
+template <typename Whole, typename LensTuple, typename PartTuple, size_t... Seq>
+decltype(auto) set_fold_impl(Whole&& whole,
+                             LensTuple&& lenses,
+                             PartTuple&& parts,
+                             std::index_sequence<Seq...>)
+{
+    using fold_op::operator+;
+    return (LAGER_FWD(whole) + ... +
```
arximboldi, line 65: This use of folding expressions is very smart, it had me smiling at the screen :)

However, it might be simpler to use recursion, so you don't have to define the `operator+` and `multiget` separatelly. (But maybe it's fine like this, I dunno.)

lager/lenses/tuple.hpp
```diff
@@ -0,0 +1,194 @@
+#pragma once
+
+#include <tuple>
+#include <type_traits>
+
+#include <lager/lenses.hpp>
+#include <lager/lenses/attr.hpp>
```
arximboldi, line 7: Nitpick: I prefer to include headers with more local to less local order. I.e. lager ones first, std ones later.

lager/lenses/tuple.hpp
```diff
@@ -0,0 +1,194 @@
...
+        LAGER_FWD(parts),
+        std::make_index_sequence<
+            std::tuple_size_v<std::remove_reference_t<LensTuple>>>());
+}
+
+// note: has to be const & to avoid potential moved from UB
+template <class T, size_t... Seq>
+auto dup_impl(T const& x, std::index_sequence<Seq...>)
+{
+    return std::tuple<decltype(static_cast<void>(Seq), x)...>(
```
arximboldi, line 84: Nice use of `static_cast<void>` for safety. Could also be `(void)x`?

lager/lenses/tuple.hpp
```diff
@@ -0,0 +1,194 @@
...
+    return explode(attr(member)...);
+}
+
+/**
+ * N -> Lens<(P1, ..., Pn), PN>
+ *
+ * Note: works for pairs, tuples, arrays. don't use on variants.
+ */
+template <size_t N>
+auto element = detail::element_t<N>{};
```
arximboldi, line 186: `inline`?

lager/lenses/tuple.hpp
```diff
@@ -0,0 +1,194 @@
...
+}
+
+/**
+ * Lens<W, P1>, ..., Lens<W, Pn> -> Lens<W, (P1, ..., Pn)>
+ *
+ * Note: parts MUST NOT OVERLAP. if they do the fold will
+ * overwrite sequential writes to the part. just don't do it.
+ */
+template <typename... Lenses>
+auto explode(Lenses&&... lenses)
```
arximboldi, line 151: Ok... so the names `zip` and `explode`, did you borrow them from somewhere?

After reading the unit tests, intuitively I feel like I would call `explode` `zip` instead. And `zip`... maybe I would call `apply` or `each` or alike?

lager/lenses/tuple.hpp
Siapran: `templated_type_constructor` is here to match against a specific `tuple` or `pair` (could be extended to arrays but I couldn't find a good use for it), and gives me a way to construct a value of that particular tuple templated type with any arguments.

lager/lenses/tuple.hpp
Siapran: I just prefer fold expressions over recursion for this kind of thing, it should improve compile times at least marginally.

lager/lenses/tuple.hpp
Siapran: yeah but a lot of tools (that includes some code reviewers) will give warnings for C style casts, even for `(void)x`.

lager/lenses/tuple.hpp
Siapran: So my train of thought was that I was zipping _lenses_ together to match the size of a tuple. `apply` would work too, though it refers really to the actual viewing/setting of the resulting lens.

`explode` was named as such because it gets you a bunch of parts for a single whole, instead of just one part as most lenses do. I don't think `zip` conveys that same meaning terribly well.

`explode` also makes it look dangerous, which it can be if you're not careful and give it lenses with overlapping parts.

Maybe we could call it `split` or `fan` or something like this? `zip` really doesn't make sense to me for this...

lager/lenses/tuple.hpp
Siapran: yep, my bad, I'll fix it.

lager/lenses/tuple.hpp
arximboldi: Well, it depends on what direction you are looking from. You can also think that you are starting with a bunch of parts, and zipping them together into a single whole ;)

lager/lenses/tuple.hpp
arximboldi: Sounds good.

lager/lenses/tuple.hpp
arximboldi: Damn tools... :p

lager/lenses/tuple.hpp
Siapran: yeah but that's also what my current `zip` does. `explode` has the benefit of conveying the fact that it expects lenses with the same single whole...

lager/lenses/tuple.hpp
Siapran: I tried rewriting this with a member templated type instead, but it doesn't really make the code any more readable or even shorter... it's just in a kind of awkward spot.

lager/lenses/tuple.hpp
arximboldi: Ok... so the other day I was writting a long response trying to argue for other naming and I eventually realized that my arguments supported... your current naming! I've also tried to look into the Haskell lens package w/o much resolution. So maybe... I'm happy to accept whatever naming is your favourite. It stills feels a bit weird this zipping when you look at it from the POV of the data that goes through (not the lenses themselves), but I don't have a better answer. Maybe just `explode` -> `fan` for the latter combinator just to avoid something too agressive? :p

arximboldi:
Cool, thanks a lot for the newer changes! I think I am happy to merge this, once you give me your final word on the naming :)

lager/lenses/tuple.hpp
Siapran: I'm ok with `fan`, it describes the shape of the lens combination pretty well.

Siapran:
alright, ready to merge!

arximboldi:
Fantastic, thank you a lot!

Issue #62, opened 2020-06-19
List of child nodes can shrink during iteration.


Siapran:
https://github.com/arximboldi/lager/blob/master/lager/detail/nodes.hpp#L185

As discussed in the past, a potential fix to this is to make the list of child nodes an immutable data structure and work on a copy. This, however, means that we are pulling Immer as a dependency for Lager, which should be fine considering Lager is designed to work best with immutable data structures, and there aren't exaclty that many implementations out there...

Siapran:
So there are two approaches to solving this:
- make `children_` an `immer` collection so `notify()` can work on a copy.
- make `garbage` into a member boolean and only clear it in `collect()`, and only call collect at the end of the outermost nested call to `notify()` (keep a `size_t notify_depth_` member to keep track of nesting depth).

I'm testing the second solution right now, and it seems to work fine, though I wouldn't trust it without going through some more rigorous testing. Using immutable collections still seems like the better idea, although performance might suffer.

arximboldi:
I'm not against the second solution. Anyways I think this might take a revamp once we integrate data diffing in, because then we potentially need to store children in a map, not vector.

PR #65, opened 2020-07-17
Fix list of child nodes shrinking during iteration (issue #62)

Siapran:
See issue #62.

I opted for the stateful approach. This fix might not be completely exhaustive, as discussed before, but it will at least stop this particular crash from happening in some projects.

We still advise against using the `lager::automatic_tag` for most cases.

### DIFF
--- a/lager/detail/nodes.hpp
+++ b/lager/detail/nodes.hpp
@@ -118,13 +118,6 @@ auto has_changed(const T&, const T&)
 template <typename T>
 class reader_node : public reader_node_base
 {
-    enum flags_t
-    {
-        no_flag              = 0,
-        needs_send_down_flag = 1 << 0,
-        needs_notify_flag    = 1 << 1,
-    };
-
 public:
     using value_type  = T;
     using signal_type = signal<const value_type&>;
@@ -155,17 +148,18 @@ public:
     void push_down(U&& value)
     {
         if (has_changed(value, current_)) {
-            current_ = std::forward<U>(value);
-            flags_ |= needs_send_down_flag;
+            current_         = std::forward<U>(value);
+            needs_send_down_ = true;
         }
     }
 
     void send_down() final
     {
         recompute();
-        if (flags_ & needs_send_down_flag) {
-            last_  = current_;
-            flags_ = needs_notify_flag;
+        if (needs_send_down_) {
+            last_            = current_;
+            needs_send_down_ = false;
+            needs_notify_    = true;
             for (auto& wchild : children_) {
                 if (auto child = wchild.lock()) {
                     child->send_down();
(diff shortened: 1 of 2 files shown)


lager/detail/nodes.hpp
```diff
@@ -178,17 +178,19 @@ class reader_node : public reader_node_base
...
         if (flags_ == needs_notify_flag) {
-            flags_ = no_flag;
+            flags_         = no_flag;
+            bool outermost = outermost_;
+            outermost_     = false;
             observers_(last_);
-            auto garbage = false;
             for (std::size_t i = 0, size = children_.size(); i < size; ++i) {
                 if (auto child = children_[i].lock()) {
                     child->notify();
```
arximboldi, line 187: This code is dangerously not exception-safe now. If `notify` throws `outermost_` will stay false forever and no later notification would succeed. Use RAII to set the flag.

lager/detail/nodes.hpp
```diff
@@ -200,17 +202,24 @@ class reader_node : public reader_node_base
...
 
-    int flags_ = no_flag;
     value_type current_;
     value_type last_;
     std::vector<std::weak_ptr<reader_node_base>> children_;
     signal_type observers_;
+
+    char flags_     = no_flag;
+    bool outermost_ = true;
+    bool garbage_   = false;
```
arximboldi, line 222: Not fan of having 3 different flag types, i'd like to keep state as small as possible. I believe that:

- `garbage_` we can just get rid of. In the worst case we would just collect garbage one notification later. It's no problem to deal with it the way we did, IMHO.
- `outermost_` can be just one more state in `flags_`.

lager/detail/nodes.hpp
Siapran: Yep, you're absolutely right. I knew something felt off when I wrote it and I want sure how to fix it at the time.

lager/detail/nodes.hpp
Siapran: so you're saying go back to just the local variable for `garbage` and only execute it for the outermost call?

arximboldi:
>
> so you're saying go back to just the local variable for `garbage` and only execute it for the outermost call?
>

yes :)

-- 
∿∿∿  https://sinusoid.al
∿∿∿  https://sinusoid.es

lager/detail/nodes.hpp
Siapran: so after reviewing all the options for flags, it feels like we don't get much benefit from packing them into a single integer, and having them as separate variables increases readability.

lager/detail/nodes.hpp
Siapran: I decided on using a `unique_ptr` with a custom deleter to do the RAII for `outermost_`.

lager/detail/nodes.hpp
```diff
@@ -155,17 +148,18 @@ class reader_node : public reader_node_base
...
     void send_down() final
     {
         recompute();
-        if (flags_ & needs_send_down_flag) {
-            last_  = current_;
-            flags_ = needs_notify_flag;
+        if (needs_send_down_) {
+            last_            = current_;
+            needs_send_down_ = false;
+            needs_notify_    = true;
```
arximboldi, line 162: So before we where setting both needs_send_down and neets_notify in one step and now in two. I do not think that this is an improvement. (This is actually moving the code back to how it was some months ago...)

lager/detail/nodes.hpp
arximboldi: While the compiler may see through the various layers of indirection, a `unique_ptr` feels a bit overkill for this...

lager/detail/nodes.hpp
Siapran: ok, but adding `outermost` to the `flags_` variable means that we now need to do some additional bitwise masking to avoid touching it.

I feel like I just need to write an extra struct to encapsulate the bitwise stuff and expose it via a more readable interface...

arximboldi: Nah ok.

lager/detail/nodes.hpp
arximboldi: Also the code is incorrect, it should set it back to `outermost`, not to `true`.

Siapran: it's not more expensive than using the `this` pointer. `unique_ptr` isn't going to allocate anything, I'm just passing it a pointer to my bool flag and a custom deleter. I thought this was the canonical way to do RAII wrapping?

PR #68, opened 2020-08-15
add missing includes

XzoRit:
When compiling with clang-9 I got this error:
```
..../lager/lager/event_loop/manual.hpp:25:31: error: expected ';' after
      expression
        throw std::logic_error{"manual_event_loop does not support async()"};
```
adding
```
#include <stdexcept>
```
in manual.hpp fixed the problem.
Then i moved on and added missing includes to all header files inside the event_loop folder :smile: .

### DIFF
--- a/lager/event_loop/boost_asio.hpp
+++ b/lager/event_loop/boost_asio.hpp
@@ -19,6 +19,7 @@
 
 #include <functional>
 #include <thread>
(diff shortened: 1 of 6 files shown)


arximboldi:
Excellent! Thank you very much!!! <3

PR #75, opened 2020-08-30
Add documentation for cursors

tusooa:


tusooa:
Also, should the "lenses" section be put before "cursors" section instead, as cursors are built upon lenses?

Siapran:
This is good. I made the lenses page a few weeks back, and we've been putting off writing the cursors documentation for too long.

Lenses are powerful tools for working with cursors, but the latter are first and foremost built on top of the Transducers API, and do not require knowledge of lenses to use, so I'm in favor of keeping them before lenses.

Siapran:
I like this, but a lot of it feels more like a general guide to writing Lager applications than a detailed documentation on cursors. `with` expressions, `.xform` and the lens syntactic sugar (the `[]` operator) are missing, when they should be the primary focus of this page.

Again, this is good content, but I feel that a lot of it would be better suited for another page.

tusooa:
> I like this, but a lot of it feels more like a general guide to writing Lager applications than a detailed documentation on cursors. `with` expressions, `.xform` and the lens syntactic sugar (the `[]` operator) are missing, when they should be the primary focus of this page.
> 
> Again, this is good content, but I feel that a lot of it would be better suited for another page.

Thank you, sure, I wrote most of the things based on the talk video referred to in the page. Will reconsider the content and change it.

doc/cursors.rst
```diff
@@ -3,16 +3,435 @@
...
+.. code-block:: c++
+
+   // Call signatures of cursors
+   model_type cursor_type<model_type>::get();
+   void lager::watch(cursor_type<model_type> cursor,
+                     std::function<void(model_type)> callback);
+   // for some types only
+   void cursor_type<model_type>::set(model_type new_model);
+
+Lager provides four types of cursors:
```
arximboldi, line 28: There are some more cursors that are not mentioned here: `lager::writer`, `lager::sensor`, `lager::constant` and "lens cursros" and "transduced cursor".

I like distinguishing between two types of cursors, "interface" cursors, "root" cursors, and "derived" cursors. `reader`, `writer` and `cursors` are the generic, interface, cursors. All other cursors are convertible to it.

Then we have "root" cursors, that are sources of data, roots of the cursor tree. These are `state`, `store`, `sensor` and `constant`.

Then we have "derived" cursors, that are created with `with(...)`, `.xform()`, `.zoom()` and `operator[]` from other cursors, forming inner and leaf nodes of the cursor tree.

doc/cursors.rst
```diff
@@ -3,16 +3,435 @@
...
+
+Transformations
+---------------
+
+The ``xform()`` function is another way to transform the cursor.
+For read-only cursors (``lager::store`` and ``lager::reader``),
+it takes one transducer; for read-write cursors (``lager::state``
+and ``lager::cursor``), it can take two to transform into another
+read-write cursor, or take one to transform into a read-only
+cursor.
```
arximboldi, line 141: This could use a reference to `https://github.com/arximboldi/zug`

arximboldi (approved):
This is great work @tusooa !!! I have some comments about some of the stuff, but overall is very good. Do you want me to merge it as is or do you want to tweak the text a bit more?

doc/cursors.rst
```diff
@@ -3,16 +3,435 @@
 Cursors
 =======
 
-.. admonition:: Work in progress
-   :class: danger
+Built upon :ref:`lenses`, cursors are a great way to bridge
+value-oriented designs and object-oriented systems. It also allows
+modularity.
```
arximboldi, line 8: "It enables modularity by allowing one to write components that reference mutable and observable data abstracting away the shape of the data storage."

Siapran (approved):

tusooa:
> 
> 
> This is great work @tusooa !!! I have some comments about some of the stuff, but overall is very good. Do you want me to merge it as is or do you want to tweak the text a bit more?

Thank you! Let me revise it a little bit.

arximboldi:
Cool, thanks a lot! <3

Wrt to your question:
> Also, should the "lenses" section be put before "cursors" section instead, as cursors are built upon lenses?

Cursors before lenses. Cursors are more important for the user and can be used without a deep understanding of lenses or transducers.

Thanks!

tusooa:
Made changes as per the suggestions.

arximboldi (approved):
This looks great! Thank you so much! :heart:

PR #82, opened 2020-11-03
Update store.hpp

gracicot:
This fixes a warning in the store header when compiling with C++20 enabled.

The `this` capture with `=` has been deprecated, so we must specify `this` manually to avoid the warning and the potential breakage in the future.


arximboldi:
LGTM! Thanks! :heart:

PR #97, opened 2021-01-25
Add a missing include in detail/nodes.hpp

jarzec:
In one of our projects @ Globus Medical I have a compilation error that `remove_if` (line 225) is an unknown identifier.


arximboldi:
Good catch! Thanks!

Issue #108, opened 2021-04-30
Improve Qt MOC compilation time


dgehri:
The Qt MOC compiler is notoriously slow when preprocessing. 

In particular, each time code includes _lager/extra/qt.hpp_, the MOC compiler spends about four seconds on precompiling it, even though the Qt relevant code is less than 100 lines.

Simply excluding non-Qt relevant code in _lager/extra/qt.hpp_ as follows would reduce MOC processing time to a fraction of a second:

```
#ifndef Q_MOC_RUN
#include <lager/cursor.hpp>
#include <lager/reader.hpp>
#include <lager/watch.hpp>
#endif
```

arximboldi:
Thanks for the fix!

PR #109, opened 2021-04-30
#108: Reduce Qt MOC compilation time

dgehri:
The Qt MOC compiler is notoriously slow when preprocessing.

In particular, each time code includes lager/extra/qt.hpp, the MOC compiler spends about four seconds on precompiling it, even though the Qt relevant code is less than 100 lines.

Simply excluding non-Qt relevant code in lager/extra/qt.hpp as follows would reduce MOC processing time to a fraction of a second:

#ifndef Q_MOC_RUN
#include <lager/cursor.hpp>
#include <lager/reader.hpp>
#include <lager/watch.hpp>
#endif

### DIFF
--- a/lager/extra/qt.hpp
+++ b/lager/extra/qt.hpp
@@ -12,9 +12,11 @@
 
 #pragma once
 
+#ifndef Q_MOC_RUN
 #include <lager/cursor.hpp>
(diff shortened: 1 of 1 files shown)


arximboldi:
Cool, thanks!

PR #111, opened 2021-05-01
Change std::result_of<> for std::invoke_result<>

Yohsi:
`std::result_of<>` has been deprecated in C++17 and removed in C++20


arximboldi:
Thank you!

PR #113, opened 2021-05-06
Use full SDL2 include path

luismerino:
Should the `SDL.h` include path be replaced with `SDL2/SDL.h`?


arximboldi:
Was this necessary in the end? It seems that the other project built correctly without this change. Happy to change it though if it doesn't really change things, I like being explicit.

luismerino:
ah! didn't notice we were already using it! nevermind then :)

PR #114, opened 2021-05-18
Add async main loop for SDL+Emscripten

luismerino:
According to https://emscripten.org/docs/porting/emscripten-runtime-environment.html#browser-main-loop, infinite main loops and functions that never return are problematic when running in the browser and they should be replaced by their async version based on `emscripten_request_animation_frame_loop()`.

This PR adds support for this through `sdl_event_loop.run_async()`.


arximboldi (approved):
Awesome!

Issue #96, opened 2021-01-21
Asynchronous action feedback / chaining


tusooa:
Assume we have some action that performs some async tasks:
```
ModelResult update(Model m, SomeAction)
{
    return { m, [](auto &&ctx) { 
        lager::get<SomeAsyncHandler>(ctx).someAsyncRequest()
            .then([=](auto res) { ctx.dispatch(UpdateModelAccordingly{res}); });
    } };
}
```

And we want to perform some other action after `UpdateModelAccordingly` has run. Here, using [post-dispatch effect](https://github.com/arximboldi/lager/issues/89) would not work as we would be processing according to the model, which it does not provide. The flow is roughly like:
```
SomeAction --| async request |~~~> UpdateModelAccordingly --> (some processing + SomeOtherAction{some data from model})
```
In traditional way it may be like
```
model.someAction()
    .then([](auto m, auto res) { m.updateModelAccordingly(res); return m; })
    .then([](auto m) { someProcessing(); m.someOtherAction(some data from model); return m; });
```
Although it is possible to use wrapper actions for that, but it is cumbersome when the flow gets longer (3-4 requests).
Another solution may be to use lambdas as actions, but then they are no longer good value types.

Is there a better solution for this? Thank you.

arximboldi:
Hi!

I do not understand why the post-dispatch effect or even the current API don't work.

Can't you do?

```cpp
ModelResult update(Model m, SomeAction)
{
    return { m, [](auto &&ctx) { 
        lager::get<SomeAsyncHandler>(ctx).someAsyncRequest()
            .then([=](auto res) { ctx.dispatch(UpdateModelAccordingly{res}); });
    } };
}

ModelResult update(Model m, UpdateModelAccordingly)
{
    // update model...
    return { m, [m](auto &&ctx) {  // <-- capture here the model or parts you need
         // more processing
         ctx.dispatch(SomeOtherAction{<some data from model>})
    } };
}
```

arximboldi:
I think it would be fair though to consider changing the API such that you know the resulting model after an action has been delivered:
```cpp
store.dispatch(some_action{})
    .then([](auto ctx, auto model) {
         // more effects depending on action having been deliverd 
    });
```
The problem is that while we can implement that for `store`, this can not be done in general for a `context` because it is agnostic of the underlying model type. Maybe you can workaround by providing a `reader` as part of the `deps` provided by the context.

arximboldi:
[...]

Issue #89, opened 2020-12-16
Add post dispatch effects (dispatch futures)


arximboldi:
The idea is that `dispatch()` may return a future that is realized when the action is actually executed. This would allow us to do stuff like:
```cpp
store.dispatch(some_action{})
    .then([](auto ctx) {
         // more effects depending on action having been deliverd 
    });
```

arximboldi:
I am working on this finally! This should help with #96 and other things.

### Current design

1. Add implementation of `promise`/`future` with monadic This design is less general than `std::future`, `boost::future` and `stlab::future`, because we use the existing notion in Lager of an event loop (something like an executor that guarantees non parallel execution of callbacks, similar to a `boost::asio::strand`).

2. `lager::context<...>::dispatch` will now return a `future<void>`. 

3. Effects can now return a `future`.

3. The `future` is resolved when the action reducer and the futures coming from the effects are fully resolved.

4. Note that a future of `void` is required because `context<>` does not know the type of the model of the store it is associated to.  If the model is required, one may as well just capture a `reader<>` in the read callback. Since the callback is guaranteed to execute in the event loop, it is safe to read there.

### Open questions

1. Is the future resolution happening too late? Are there super significant cases where one may want to chain to an earlier point? (e.g. reducer has run, but not effects.)

2. This may be an API breaking change and introduce overhead, should it be optional?

3. To be honest, it seems like we are gonna be using mostly `future<void>` should that the only thing that exists in the library, which greatly simplifies things?

4. Considering the restrictions, particularly if we only provide `future<void>` maybe the name `future` is a misleading name, what other names could we use? Maybe `token` or `operation` or alike?

---

Looking for feedback. Ping @tusooa @Siapran.

PR #128, opened 2022-01-10
Add basic futures to observe dispatch of actions and effects

arximboldi:
This PR addresses #89 and potentially #96.


arximboldi:
Merging this now.

In a private repo, @isherman has shown that enabling futures can have a performance penalty per-dispatch. The penalty is small enough that it is probably ok for most projects, but it is still worth having it optional. So to enable it, you need to pass the `lager::with_futures` enhancer to the `make_store` call. In the future I am gonna explore other designs that either remove the performance penalty or make it optional per `dispatch()` instead of per store.

PR #133, opened 2022-02-10
Fix unpacking the of nodes in lens reader nodes

dimula73:
It breaks applying a lens expression over the result of .with() expression.

### DIFF
--- a/test/lenses.cpp
+++ b/test/lenses.cpp
@@ -16,6 +16,7 @@
 #include <zug/compose.hpp>
 #include <zug/util.hpp>
 
+#include <lager/state.hpp>
 #include <lager/lenses.hpp>
 #include <lager/lenses/at.hpp>
 #include <lager/lenses/at_or.hpp>
@@ -609,4 +610,22 @@ TEST_CASE("lenses::element array", "[lenses][element][array]")
     CHECK(view(second, foo) == 2);
 
     CHECK(over(element<1>, foo, increment) == (std::array{1, 3, 3}));
+}
+
+TEST_CASE("lenses over with expression")
+{
+    state<person, automatic_tag> person_data;
+
+    person_data.set(person{{}, "old name", {}});
+
+    auto name_lens = lenses::getset(
+        [] (std::tuple<std::string, yearday> x) { return std::get<0>(x); },
+        [] (std::tuple<std::string, yearday> x, std::string new_name) { return std::make_tuple(new_name, std::get<1>(x)); });
+
+    cursor<std::string> name = with(person_data[&person::name], person_data[&person::birthday]).zoom(name_lens);
+
+    name.set("new name");
+
+    CHECK(person_data->name == "new name");
+    CHECK(name.get() == "new name");
 }
--- a/test/lenses.cpp
+++ b/test/lenses.cpp
@@ -16,7 +16,6 @@
 #include <zug/compose.hpp>
 #include <zug/util.hpp>
 
-#include <lager/state.hpp>
 #include <lager/lenses.hpp>
 #include <lager/lenses/at.hpp>
 #include <lager/lenses/at_or.hpp>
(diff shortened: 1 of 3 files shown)


arximboldi:
Oh wow, great catch! Clearly I mostly use `.map(...)` after `with()` expression :facepalm: It'd be great if you added to the PR, but if you don't have the time let me know and I will add it after the merge.

dimula73:
Hi, @arximboldi!

I'm not sure about what you want to be added to the PR :) I have briefly grepped the source code for the constructions like that, and it seems like xform and merge node don't have this misprint.

And I couldn't use the `.xform()`/`.map()` expression, because I need a reduction from two source values into one and I cannot define the reverse transformation for them.

```
auto brushSizeLens = lager::lenses::getset(
    [](std::tuple<QSize, qreal> x) -> qreal { return std::get<0>(x).width() * std::get<1>(x); },
    [](std::tuple<QSize, qreal> x, qreal brushSize) -> std::tuple<QSize, qreal> {
        return std::make_tuple(std::get<0>(x), brushSize / std::get<0>(x).width());
    });
```

PS:
Btw, just for the context, right now I'm experimenting with using lager in Krita opensource painting application :)
https://invent.kde.org/dkazakov/krita/-/commit/f33883b0e2a8609f41ac6e3dbefa01ea07991c46#d8b2abffd9405d4e7af577135ca9fb02bfbf6b18_36_48

arximboldi:
Sorry, I wrote the comment before having a coffee so I skipped half of the words.

I meant: "It'd be great if you added a *unit test* to the PR", covering the case you already covered with the fix (I am aware that this problem does not exist for `with` combined `xform()`, because that's precisely the pattern I normally use and the oldest version of this construction.) Do you want to add the unit test or should I do it?

In any case, thank you so much for the contribution, and it makes me very happy that you are experimenting using it in Krita! As a Free Software advocate and Krita user myself, it makes me very happy, Krita is definitely one of the best open-source tools for creators out there :slightly_smiling_face:  :heart:

dimula73:
> I meant: "It'd be great if you added a unit test to the PR"

I can try to, but I will probably need so guidance. And it will not happen today. I'm on Windows right now and there is some test configuration issue with unittests :( Is the library actually tested on Windows?

Should the testcase go into test/cursor.cpp? Or something else?

> For completenes: you can also use map for that, by passing a second mapping function that maps the target value into a tuple.

Erm... the problem is that I cannot calculate the source value from the target value without knowing original source value. The function is not injective.

dimula73:
>  Krita is definitely one of the best open-source tools for creators out there 🙂 ❤️

And thanks a lot! :)

arximboldi:
> Erm... the problem is that I cannot calculate the source value from the target value without knowing original source value. The function is not injective.

Oh yeah, there is a dirty trick I added at some point before lenses existed, see `lager::update()` in `with.hpp`, it's a transducer for that case... Anyway, your solution is cleaner :smile:

arximboldi:
> I can try to, but I will probably need some guidance. And it will not happen today. I'm on Windows right now and there is some test configuration issue with unittests :( Is the library actually tested on Windows?

Since rencently the unit tests should work on Windows. The tests are however not run on Windows CI.

> test/cursor.cpp

Maybe `test/xformed.cpp`

dimula73:
Hi, @arximboldi!

I have added the unittest into `lenses.cpp`, because it already had a structure that could be used for testing. Though it added 'state.hpp' include into this test. I'm not sure that is okay.

test/lenses.cpp
```diff
@@ -610,3 +611,21 @@ TEST_CASE("lenses::element array", "[lenses][element][array]")
...
     CHECK(over(element<1>, foo, increment) == (std::array{1, 3, 3}));
 }
+
+TEST_CASE("lenses over with expression")
+{
+    state<person, automatic_tag> person_data;
+
+    person_data.set(person{{}, "old name", {}});
+
+    auto name_lens = lenses::getset(
```
Siapran, line 621: Could you replace this lens with `lager::lenses::first` from tuple.hpp?

Siapran:
@dimula73, you could move the new test into [cursor.cpp](https://github.com/arximboldi/lager/blob/master/test/cursor.cpp), the `person` structure is also present there.

arximboldi:
This is great, thank you so much for taking the time to add the unit tests!

PR #134, opened 2022-02-10
Fix unittests when Zug and Immer are installed outside system prefix

dimula73:


arximboldi:
Sadly the the casing change is breaking the build in Linux.

dimula73:
Hi, @arximboldi!

I've just pushed a fix :)

arximboldi (changes requested):
Sadly the the casing change is breaking the build in Linux. Are you sure it's needed. The "module" uses CamelCase, but then the target uses lower-case, I thought that's CMake convention.

arximboldi (changes requested):
Now the `build` run is failing, to can be fixed by adding:
```
deps.zug
```
to `propagateBuildInputs` in line 31 to `default.nix`.

Sorry about that!

dimula73:
Can it be that `CMAKE_PREFIX_PATH` is not set properly on the CI? At least here I have `ZugConfig.cmake` in the install directory of zug: `${CMAKE_INSTALL_PREFIX}/lib/cmake/Zug/ZugConfig.cmake`

dimula73:
Btw, @arximboldi, do you have any IRC channel or something? I wanted to ask about an unrelated issue I have with lager::state not notifying observers, even though lager::cursor created from the same state works fine. I'm not yet sure it is a bug in lager or not, because I cannot reproduce that in a unittest (it works fine in a unittest). So I would really appreciate a hint into which direction where I should look into.

arximboldi:
@dimula73 you can send me an email with example code, or even open an issue here tagged as `question`.

PR #136, opened 2022-02-14
Delete ambiguous assigment operators in `state`

### LINKED ISSUE #135: Issue with notifications from lager::state
Hi!

I have a weird issue with notifications from lager::state<>. For some reason, the state doesn't emit any notifications, but if I wrap it into a cursor, then the notifications work fine. If I do `state.nudge()`, then the notifications arrive fine, but when I change the contents of the state using cursors to its members, the state keeps silence.

I tries to reproduce the issue in a unittest, but failed. In unittest it works fine, but not in my real code. I'm still not sure this bug is in lager itself or in my code. Could you suggest me any direction, where I should look at? Or at least how to debug the issue?

```cpp
#include <boost/operators.hpp>

namespace testns {
struct Child : public boost::equality_comparable<Child>
{
    friend bool operator==(const Child &lhs, const Child &rhs) {
        return lhs.childValue == rhs.childValue;
    }

    Child() = default;
    Child(int value) : childValue(value) {}

    int childValue = 0;
};

struct Parent : public boost::equality_comparable<Parent>
{
    friend bool operator==(const Parent &lhs, const Parent &rhs) {
        return lhs.parentValue == rhs.parentValue &&
            lhs.child == rhs.child;
    }

    int parentValue = 0;
    Child child;
};
}

TEST_CASE("state, check notifications come after edits through a cursor")
{

    lager::state<testns::Parent, automatic_tag> state;
    lager::cursor<testns::Parent> intermCursor = state;
    lager::cursor<testns::Child> childCursor = intermCursor[&testns::Parent::child];
    lager::cursor<int> childValueCursor = childCursor[&testns::Child::childValue];

    auto childSpy = testing::spy([&](const testns::Child &value) {
        CHECK(value.childValue == 13);
    });

    auto parentSpy = testing::spy([&](const testns::Parent &value) {
        CHECK(value.parentValue == 0);
        CHECK(value.child.childValue == 13);
    });

    watch(state, parentSpy);
    watch(childCursor, childSpy);

    //childCursor.set(testns::Child(13));
    childValueCursor.set(13);

    CHECK(state.get().parentValue == 0);
    CHECK(state.get().child.childValue == 13);

    CHECK(childSpy.count() == 1);

    /// HERE!!! In real code, parent doesn't emit anything, but in this unittest the signal works fine
    CHECK(parentSpy.count() == 1);
}

```

dimula73:
It is possible to assign a state with the new value, which implicitly
creates a new state and replaces the current one. It basically detaches
all the cursors and listeners.

Alternatively, the problem can be solved by making the constructor
explicit, but it may create some inconviniences when using the state.

fixes #135

### DIFF
--- a/lager/state.hpp
+++ b/lager/state.hpp
@@ -94,6 +94,13 @@ public:
 
     state(state&&) = default;
     state& operator=(state&&) = default;
+
+    /*!
+     * Deletes implicit creation and replacement of the store
+     * when it is assigned with the base type
+     */
+    state& operator=(const T&) = delete;
+    state& operator=(T&&) = delete;
 };
 
 template <typename T>
--- a/test/state.cpp
+++ b/test/state.cpp
@@ -10,6 +10,8 @@
 // or here: <https://github.com/arximboldi/lager/blob/master/LICENSE>
 //
 
+#include <boost/operators.hpp>
+
 #include <catch.hpp>
 
 #include <lager/state.hpp>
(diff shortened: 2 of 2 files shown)


arximboldi (approved):
LGTM! :+1: Thank you for taking care of this!

PR #141, opened 2022-03-30
Forward reducer in with_reducer enhancer

Siapran:
This was causing a weird binding error in clang on the `reducer` capture of the inner lambda when the enhancer is passed an lvalue.


arximboldi (approved):
Awesome! Thanks Siapran! <3

PR #142, opened 2022-04-06
Implement LAGER_DERIVE_SIZE_CHECK

jcelerier:


jcelerier:
@arximboldi I wanted to add a test, but a proper test is "check that LAGER_STRUCT(...) does not compile", how would you go about this ? 
also, for the template case it's definitely not implementable (as it's only meaningful for a specific instantiation) but for the nested case I'm not sure.. but my boost-pp-fu isn't high enough aha

Siapran:
this is nice, but you could replace the `boost::pfr::tuple_size_v<name__>` by a structured binding, which would allow you to implement the size check for the template and need versions too

Siapran:
well yeah it'll only error on instantiation. the main use I'd have for it is to yell at me when I forget to update the members in the macro when I add something to the type, which means at that point it'll have been instantiated where used already.

Siapran:
as for readable error messages you can probably SFINAE it now that I think about it.

jcelerier:
> well yeah it'll only error on instantiation

so, I'm not super familiar with the library - but instantiation of what ? just instantiating the type won't instantiate e.g. a separate 

     template<typename T>
     void lager_check_struct_size() { 
       auto [a1,a2,a3,...] = ns::my_type<T>{};
     }

unlike the normal struct cae

Siapran:
oh you're right, my bad. I made my own version a while back but baked it into the cereal functions so it'd be instantiated for serialization.

jcelerier:
Argh, I tried to implement a version with structured bindings, but sadly I can't make it work: the issue is the zero case: 

     auto [ ] = something;

is not even valid syntax so we can't hide between SFINAE, if constexpr or anything like that when expanding __members

jcelerier:
although.. maybe with some BOOST_PP_IF..

arximboldi:
How should we go about this? Do you want to add some kind of test or should I just merge as is? Happy to merge as it is to be honest...

PR #143, opened 2022-04-08
Add const qualification to HASH implementation.

Siapran:
This caused issues when attempting to use the generated hashes with standard containers.


arximboldi:
Good catch, thank you!

Issue #140, opened 2022-02-24
Reimplement the lager::http_debug_server using Boost.Beast


arximboldi:
Currently, the debugger depends on [libhttpserver](https://github.com/etr/libhttpserver). This is the only dependency other than `zug` and `beast`. While this is optional, and only required when `lager/debug/http_server.hpp` is included, this is an ackward depedency that limits the widespread of the debugger. Before going on further polishing of the debugger, whose code is getting a bit old (and there are some quircks when combining it with newer functionality like cursors as highlighted by #139 and #137), we should clean this up and just reimplement it using `boost::beast`.

Beast is a bit lower level, so it may require a bit of extra boilerplate, but the whole server is small enough that it shouldn't be too difficult. This is a rather self-contained class also, without too heavy template metaprogramming, so this may be a good task for a new contributor to the project to tackle. If someone wants do it, please let me know and I can offer further assistance!

arximboldi:
Thank you @ashtum for contributing this! :heart:

PR #144, opened 2022-05-06
Reimplement the lager::http_debug_server using Boost.Beast

ashtum:
This is a draft PR for #140

- Because lager is header only, a low footprint implementation that meets current requirements has been in mind.
- There is no dependency on [libhttpserver](https://github.com/etr/libhttpserver) anymore but cmake and nix files need modification.
- If windows support is required, manipulation of `resources_path` should be different.
- Currently, there are no unit tests for http_debug_server, we need some idea of how it should be done.

### DIFF
--- a/lager/debug/http_server.hpp
+++ b/lager/debug/http_server.hpp
@@ -20,26 +20,234 @@
 #include <cereal/types/optional.hpp>
 #include <lager/extra/cereal/variant_with_name.hpp>
 
-#include <httpserver.hpp>
+#include <boost/asio/ip/tcp.hpp>
+#include <boost/beast/core.hpp>
+#include <boost/beast/http.hpp>
 
 #include <atomic>
+#include <fstream>
 #include <functional>
 #include <memory>
+#include <regex>
 #include <sstream>
+#include <thread>
 #include <vector>
 
 namespace lager {
 
-namespace detail {
+namespace beast = boost::beast;
+namespace asio  = boost::asio;
 
-bool ends_with(const std::string& str, const std::string& ending)
+namespace detail::http {
+class router
 {
-    return str.length() >= ending.length() &&
-           str.compare(
-               str.length() - ending.length(), ending.length(), ending) == 0;
-}
+public:
+    using req_t = beast::http::request<beast::http::string_body>;
+    using res_t = beast::http::response<beast::http::string_body>;
 
-} // namespace detail
+private:
+    struct resource_t
+    {
+        beast::http::verb method;
+        std::regex target;
+        std::function<res_t(req_t&&)> handler;
+    };
+
+    std::vector<resource_t> resources_;
+
+public:
+    void register_resource(beast::http::verb method,
+                           const std::string& target,
+                           std::function<res_t(req_t&&)> handler)
+    {
+        resources_.push_back({method, std::regex{target}, std::move(handler)});
+    }
+
+    res_t handle_request(req_t&& req)
+    {
+        auto it = std::find_if(resources_.begin(),
+                               resources_.end(),
+                               [&](const resource_t& resource) {
+                                   return resource.method == req.method() &&
+                                          std::regex_match(req.target().begin(),
+                                                           req.target().end(),
+                                                           resource.target);
+                               });
+
+        if (it != resources_.end()) {
+            try {
+                return it->handler(std::move(req));
+            } catch (const std::exception& err) {
+                return create_response_(500, "text/html", err.what());
+            }
+        }
+
+        return create_response_(404, "text/html", "Not found");
+    }
+
+private:
+    static res_t create_response_(unsigned status_code,
+                                  beast::string_view content_type,
+                                  std::string body)
+    {
+        res_t res;
+        res.result(status_code);
+        res.set(beast::http::field::content_type, content_type);
+        res.body() = std::move(body);
+        res.prepare_payload();
+        return res;
+    }
+};
+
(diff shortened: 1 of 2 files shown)


arximboldi:
This is super awesome! I'm gonna try take a deeper look over the weekend or early next week. Thanks a lot! <3

TheCoconutChef:
Is there any concern with respect to the use of `std::regex`?

Consider [this](https://www.reddit.com/r/cpp/comments/e16s1m/comment/f8offl4/?utm_source=share&utm_medium=web2x&context=3) post.

Should `Boost.regex` be considered instead? A quick look at [this recent regex benchmark project](https://github.com/mariomka/regex-benchmark) would seem to suggest so.

ashtum:
> Is there any concern with respect to the use of `std::regex`?
> 
> Consider [this](https://www.reddit.com/r/cpp/comments/e16s1m/comment/f8offl4/?utm_source=share&utm_medium=web2x&context=3) post.
> 
> Should `Boost.regex` be considered instead? A quick look at [this recent regex benchmark project](https://github.com/mariomka/regex-benchmark) would seem to suggest so.

Yes, it can be replaced with boost version. but I think for this part of code performance of regex shouldn't be an issue because the overall operation took orders of magnitude longer than that regex and this is not an under load http server.

Siapran:
Well one of the issues with `std::regex` is that even just including the header is detrimental to compile times, which in turn would affect CI.

ashtum:
> Well one of the issues with `std::regex` is that even just including the header is detrimental to compile times, which in turn would affect CI.

I couldn't measure any compile time difference, maybe that is because asio and beast got a large portion of compile time.
I will be glad if you measure the difference and it's impact on CI.
And Boost.regex needs a linkage to lboost_regex.

Siapran:
Actually nvm, you're right: boost regex takes longer to compile than std regex.
I was misremembering the conclusion to [this talk](https://youtu.be/8dKWdJzPwHw) by Hana Dusíková, where she presents her own regex library.

arximboldi:
I'd go with @ashtum here and prefer the option that does not require special linker flags

lager/debug/http_server.hpp
```diff
@@ -100,128 +308,158 @@ class http_debug_server
...
+        static beast::string_view mime_type_(beast::string_view path)
+        {
+            using beast::iequals;
+            auto const ext = [&path] {
+                auto const pos = path.rfind(".");
+                if (pos == beast::string_view::npos)
+                    return beast::string_view{};
+                return path.substr(pos);
+            }();
+            if (iequals(ext, ".htm"))
```
arximboldi, line 407: An alternative way to do this would be to convert the `ext` to lowercase and then replace all these `if/else` chain with a lookup in a dictionary (`std::unordered_map` or `immer::map`).

lager/debug/http_server.hpp
```diff
@@ -100,128 +308,158 @@ class http_debug_server
...
+            std::ifstream ifs{full_path};
+            if (!ifs.is_open())
+                return create_response_(404, "text/html", "Not found");
+
+            /*
+              The whole file content is read synchronously into string, this
+              shouldn't be a performance issue because it's used only in gui
+              startup and has no blocking effect on the application because
+              http::server has its own thread.
+            */
```
arximboldi, line 392: Nitpick: I prefer `//` style comments for C++ code, particularly for comments in the code (use `/*! ... */` for Doxygen comments.)

lager/debug/http_server.hpp
```diff
@@ -230,38 +468,78 @@ class http_debug_server
...
-        server_.register_resource("/api/step/{cursor}", &hdl.step_resource_);
-        server_.register_resource("/api/goto/{cursor}", &hdl.goto_resource_);
-        server_.register_resource("/api/undo", &hdl.undo_resource_);
-        server_.register_resource("/api/redo", &hdl.redo_resource_);
-        server_.register_resource("/api/pause", &hdl.pause_resource_);
-        server_.register_resource("/api/resume", &hdl.resume_resource_);
-        server_.register_resource("/api/?", &hdl.root_resource_);
-        server_.register_resource("/?.*", &hdl.gui_resource_);
+
+        detail::http::router router;
```
arximboldi, line 485: Nitpick: I prefer AAA-style: https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/

lager/debug/http_server.hpp
```diff
@@ -230,38 +468,78 @@ class http_debug_server
...
+
+        namespace ph = std::placeholders;
+
+        router.register_resource(
+            beast::http::verb::get,
+            "/api/?",
+            std::bind(&handle_t::root_resource_, &hdl, ph::_1));
```
arximboldi, line 492: I think part of this design, with the resources being methods of the handle, and the handle separate from the server, was motivated by the way the old http server library worked, and it could be made more simple or elegant now... Anyway this is not a requirement to change this, just food for thought.

arximboldi (approved):
Hi!

This is super nice, thank you! Removing that dependency is something that I had long been wishing for... I have tried this and it seems to work pretty nice!

One note: there are still references to `libhttpserver` in the `shell.nix`, `default.nix` and in the CMake code. All those references can be removed now.

Anyway, I am in spite of the comments, I am happy with the contribution as it is, and I can do the final polishing myself. Let me know if you want me to merge it now, or if you want to spend more time on it before the merge.

Cheers!

PR #146, opened 2022-05-10
Fix sfinae and make library build with C++20

gracicot:
First thing, the sfinae check was not right. It was checking if the function was callable using `std::result_of_t<Fn(context_t)>`. This is not the right check, since the callable type will be stored in a `std::function<future(const context<Action, Deps>&)>` aka `std::function<future(const context_t&)>`.

This check will reject completely valid code.

A `std::function` will forward exactly a `const context_t&`, and the sfinae check should reflect exactly that to not cause unnecessary instantiations and also to make generic lambdas deduce the right type.

Here's an example:

```c++
struct context_t {};

auto lambda = [](auto&) {};

// Error, `auto&` deduce to `context_t` which makes it `(context_t&)`, does not bind to rvalues
std::result_of_t<decltype(lambda)(context_t)>;

// Okay, `auto&` deduce `const context_t` which result in `(const context_t&)`, binds to  l and r values
std::result_of_t<decltype(lambda)(const context_t&)>;
```

So the correct check would be `std::result_of_t<Fn&(const context_t&)>` so it instantiate the same thing as the `std::function` type will do.

However, I also changed the check to use simple `decltype` instead of that check, since `std::result_of` deprecated in C++17 and removed in C++20. This pull request brings back C++20 support.


gracicot:
The sfinae syntax could be factored into an alias or a custom type trait. If this is what you want let me know and I'll change the patch

gracicot:
Quick question: Can we use C++17? In that case using `std::invoke_result_t` would be more appropriated

Siapran:
Lager depends explicitly in C++17. I was going to suggest `invoke_result` earlier but I got distracted.

arximboldi:
I agree that `invoke_result` would be cleaner, would you mind applying that change @gracicot ?

arximboldi (approved):
Neat, thank you!

PR #150, opened 2022-05-30
Make `lager::match` work with types that are derived from `std::variant`

arximboldi:
One of the problems of `std::variant` is that it expands all thea lternatives in the type name.  This can lead to humongous compilation errors.  This is particularly true for the patterns encouraged by Lager, where actions are variants of variants of variants...

One solution to erase the alternatives from the type of the variant is to do something of the form:
```c++
  using variant_t = std::variant<...>;
  struct my_variant : variant_t { using variant_t::variant_t; };
```

However, `std::visit` does not work with variants of that form in some implemnetations.  [This is definitelly true for libstdc++](
https://stackoverflow.com/questions/51309467/using-stdvisit-on-a-class-inheriting-from-stdvariant-libstdc-vs-libc).

This PR works around it in lager::match itself. Also some utilities are provided that can come handy when applying the workaround when metaprograming.

### DIFF
--- a/lager/util.hpp
+++ b/lager/util.hpp
@@ -12,6 +12,7 @@
 
 #pragma once
 
+#include <zug/detail/copy_traits.hpp>
 #include <zug/util.hpp>
 
 #include <functional>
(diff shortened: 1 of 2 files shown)


arximboldi:
FYI @wtholliday

PR #161, opened 2022-10-13
Test implementation of deferred calculation of the 'get()' part of le…

dimula73:
…nses

WARNING: this patch is only a proof of concept. It breaks a lot of
         unittests and I don't know how to fix them.

The perpose of this patch is twofold:

1) Avoid execution of the 'get()' part fo the lens when only set is
   requested
2) Avoid double expiring of the object when 'get()' part of the lens
   accepts the "whole" object as a non-reference


dimula73:
This PR is now superseded by https://github.com/arximboldi/lager/pull/165

PR #162, opened 2022-10-21
Nicer build failure messages on bad assignments

TheCoconutChef:
Lager offers a lot of assignment and implicit conversion facilities. However, in many cases, bad assignment operations are not catched by language servers and the compilation error messages can be rather opaque, sometimes failing to point out the offending line to the user. The following examples are illustrative:

https://godbolt.org/z/cGMsY7Ghq

This PR aims to improve the situation through the use of sfinae on certain implicit conversion operators and chosen ctors.

Under the proposed changes, a bad assignment statement compiling in GCC will produce something like:

```
/home/foouser/project/lager/test/compilation/map_to_reader.cpp: In function ‘int main()’:                                                                                       
/home/foouser/project/lager/test/compilation/map_to_reader.cpp:12:41: error: conversion from ‘lager::detail::with_xform_expr<zug::composed<zug::map_t<main()::<lambda(auto:137)>, lager::detail::store_node_base<ModelAction, Model> >’ to non-scalar type ‘lager::reader<std::__cxx11::basic_string<char> >’ requested                                     
   12 |     lager::reader<std::string> a = s.map([](auto) { return 1; }); // offending line                                                                                      
      |                                    ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~                                                                                                         
make[3]: *** [test/compilation/CMakeFiles/test_compilation_map_to_reader.dir/build.make:76: test/compilation/CMakeFiles/test_compilation_map_to_reader.dir/map_to_reader.cpp.o] Error 1                                                                                                                                                                              
make[2]: *** [CMakeFiles/Makefile2:1293: test/compilation/CMakeFiles/test_compilation_map_to_reader.dir/all] Error 2                                                                
make[1]: *** [CMakeFiles/Makefile2:1300: test/compilation/CMakeFiles/test_compilation_map_to_reader.dir/rule] Error 2                                                               
make: *** [Makefile:699: test_compilation_map_to_reader] Error 2
```
which can be contrasted with what one can observe in the provided godbolt snippet.

Thus, this PR operates in the spirit of 43b5b59a2bae56e1770e71e5f8feec79758de8eb, i.e. trying to improve the user compile time experience. It covers the following assignments:

* cursor to cursor 
* cursor to reader
* lense to cursor
* lense to reader
* map to reader
* reader to reader
* with expr to reader
* xform to cursor
* xform to reader
 
Notably absent are the `writer`s and the constructors taking a `std::shared_ptr<NodeT2>`. In the case of the `writer`, I wasn't sure if it was lager's intention to have this type ever be used directly by the user, and so I didn't make any change. In the `std::shared_ptr<NodeT2>` case, the reason is similar: I don't know under which conditions this is used.

If other cases can / could be tested, I'd be interested to know.

**N.B. About the "compilation test"**

I have some strong doubt about the legitimacy of trying to add tests for this improvement, which is why the last commit is WIP. Testing compile error output is hard and probably not very robust. It's unclear whether or not the code or the particular compiler or compiler version is being tested. Furthermore, trying to test the compile error output has had the following disadvantages:

1. I had to bring in another dependency to run the tests, i.e. shunit2.
2. The cmake to generate the compilation tests isn't very nice.
3. I broke the glob pattern in the base `test/` dir as the compilation test cpp files shouldn't be added to the unit test target. Should  a different test directory for compilation test be created?
4. The compilation tests are slow relative to the unit tests
5. If these tests use a bash runner like shunit2, they obviously won't be that portable

Therefore, I'm pretty sure we're gonna have to drop this commit or at least modify it somewhat. However, I though including it in the PR could be interesting anyway just in case I was wrong in my pessimism and to show what was tested.

The test format consists in putting a "happy path line" and an "offending line" in a `main()` function, then run the compiler on the cpp file. We then make sure the happy path line isn't in the error output but that the offending line is. I also put a (rather brittle!) test checking that the error output is less than 10 lines long.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,7 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
 endif()
 
 option(lager_BUILD_TESTS "Build tests" ON)
+option(lager_BUILD_FAILURE_TESTS "Build failure tests" ON)
 option(lager_BUILD_EXAMPLES "Build examples" ON)
 option(lager_BUILD_DEBUGGER_EXAMPLES "Build examples that showcase the web based debugger" ON)
 option(lager_BUILD_DOCS "Build docs" ON)
--- a/cmake/icm_build_failure_testing.cmake
+++ b/cmake/icm_build_failure_testing.cmake
@@ -1,0 +1,232 @@
+# icm_build_failure_testing
+#
+# SPDX-License-Identifier: MIT
+# MIT License:
+# Copyright (c) 2022 Borislav Stanimirov
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files(the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and / or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
+# the following conditions :
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT.IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+#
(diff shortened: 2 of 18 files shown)


TheCoconutChef:
Interestingly, just came across this article: https://ibob.bg/blog/2022/10/04/testing-build-failure-with-cmake/ which could lead to the removal of the shunit2 dependency and some cmake ugliness.

TheCoconutChef:
Only tentatively. I did not have the time to investigate the issue.

Also, since one of the problem seems to be that the test cannot find the source file to compile for the build failure test, I'm not even sure the MacOS failure is actually a failure of the test assertion. Does it also fail to find the source file?

What I plan to try next is to use https://ibob.bg/blog/2022/10/04/testing-build-failure-with-cmake/ instead of shunit2. This will remove the shunit2 dependency and clean up my cmake.

If it works, great. If not, I think dropping the compilation test all together wouldn't be so bad.

arximboldi:
Cool! Yes I think using that technique should work, and otherwise also happy to drop the build-failure tests. I already had cases where they would have come in handy, but was too lazy to get it to work so I just left a commented out line in an unit test, that one can manually check from time to time.

TheCoconutChef:
Round 2. 

The use of `icm_build_failure_test` greatly simplifies the cmake. However, we lose the "happy path" negative match test and the output size test.

arximboldi (approved):
I like this! Thanks a lot for these improvements.

The last commit is marked as WIP. Do you plan to improve it or should I merge as-is?

arximboldi:
Thanks a lot @TheCoconutChef !

Issue #163, opened 2022-11-13
Crashes on access to uninitialized reader/writer


jarzec:
Access to uninitialized `lager::reader` or `lager::writer` results in unprotected access to a `nullptr` node, which terminates the program. The node pointer should be checked and an exception should be thrown in case of `nullptr`.

jarzec:
Solved

PR #164, opened 2022-11-14
Throw on access to uninitialized reader/writer

jarzec:
Fixes issue #163


arximboldi (approved):
LGTM! Thanks!

JarekGlobus:
@arximboldi Do you think there are other places this kind of mod would be useful? If not I propose to merge as-is.

arximboldi:
No, I think this is fine.

PR #171, opened 2022-12-27
Implement watchable_base::unbind()

dimula73:
It turned out that there was no way to unbind anything from a reader after once bound. Assignment operator keeps all the binding untouched (which looks a bit strange, but "by design"), so we need a separate method to unbind everything when needed.


arximboldi:
LGTM.

Agree current connection semantics, while convenient for some use-cases, have some surprising cases. Been thinking about changing them for a while, to something a bit more explicit when it comes to lifetimes.

PR #178, opened 2023-05-24
Impl. an "at setter" that works with immer::table

colugomusic:
This allows us to create a cursor to an `immer::table` element i.e.
```c++
struct item { int id; }
struct model {
  immer::table<item> items;
};
...
lager::reader<item> item_123 = store[&model::items][123].zoom(lager::lenses::or_default);
```
This is currently a compile error because `immer::table` has no `set` operation, so the first implementation of `at_setter_impl` would be selected leading to the invalid assignment.

### DIFF
--- a/lager/lenses/at.hpp
+++ b/lager/lenses/at.hpp
@@ -14,16 +14,22 @@ namespace lager {
 namespace lenses {
 namespace detail {
 
-// detect if T satifsies the immer API for setting values
+// detect if T satisfies the immer API for setting values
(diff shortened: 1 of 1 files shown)


arximboldi:
CI is failing because of outdated actions. Can you rebase on latest master so we can see a green build/check?

arximboldi:
Other than that, thank you a lot for this contribution, this is very useful... `table` was introduced relatively recently and Lager hadn't been updated to match yet.

PR #183, opened 2023-06-20
fix minor typos modularity.rst

mjcaisse:
fix minor typos modularity.rst


arximboldi:
So you're using Lager in a new project? :smile:

mjcaisse:
> So you're using Lager in a new project? smile

A side project streaming lots of data from multiple sources. Initially I was looking at zug and then remembered you have this entirely wonderful stack of immer, zug, and lager!

PR #186, opened 2023-07-30
Use Catch2 as installed

pinotree:
When building the tests, require and use Catch2 as installed in the system, rather than using an old embedded copy of it. This way, there is no risk it will stop working with newer versions of glibc/gcc/etc.

This also changes the nix packaging to disable tests & examples: they are not used in any way, so better disable them to avoid requiring bits only needed for them.


pinotree:
hmm the `build` job fails, and it seems also failed after the last push; let's see whether #187 fixes that

PR #187, opened 2023-07-30
nix: fix build of dependencies

pinotree:
- fix the `cmakeFlags` specification for cereal, as it must be a string (and not a simple string)
- build immer and zug explicitly disabling tests and examples, as they are neither run nor installed


arximboldi:
Nice one! Will merge once the tests run.

PR #189, opened 2023-08-03
cmake: use GNUInstallDirs

pinotree:
Use the standard GNUInstallDirs CMake module to define the paths where to install bits; this makes them following the OS/distro defaults, and it is possible for users to customize them.


arximboldi:
Thank you for all the polishing in Lager, Immer and Zug! :+1:

PR #190, opened 2023-08-03
Update the SDL2 CMake modules

pinotree:
- drop `FindSDL2.cmake`, since SDL2 ships config files for CMake already
- drop `FindSDL2_image.cmake`, as SDL2_image is not used
- update the license text of `FindSDL2_ttf.cmake`


pinotree:
Needs more testing -> draft for now.

arximboldi:
It seems the build passed, what kind of further testing would you like to have?

pinotree:
> what kind of further testing would you like to have?

I was not sure yet about the approach to do for `FindSDL2_ttf.cmake`, since the main problem I wanted to solve was the lack of the license text (mentioned there); while the original idea was also to provide a newer/better version of it, in the end I chose to simply fix the license text, and keep the existing module (which works fine).

Hence, it should be fine to review now, thanks!

arximboldi:
Thank you!!

PR #191, opened 2023-08-03
cmake: fix finding cursesw

pinotree:
The right variable for requesting the unicode version of curses is `CURSES_NEED_WIDE`, not `CMAKE_CURSES_NEED_WIDE` [1].

[1] https://cmake.org/cmake/help/latest/module/FindCurses.html


arximboldi:
Oh, thank you!

PR #196, opened 2023-10-06
Fix build of http_debug_server with boost >= 1.80

pinotree:
Boost Beast 1.80 switches from `boost::string_view` to `boost::core::string_view` [1][2], and thus the `to_string()` method to convert to `std::string` is not available anymore.

Both the old and the new `string_view` types used support converting to `std::string` via operator, hence switch to a `static_cast` to do the conversion. This supports building with all the versions of boost.

[1] https://www.boost.org/doc/libs/release/libs/beast/doc/html/beast/release_notes.html
[2] https://github.com/boostorg/beast/issues/2417

### DIFF
--- a/lager/debug/http_server.hpp
+++ b/lager/debug/http_server.hpp
@@ -284,7 +284,7 @@ struct http_debug_server_impl
                        auto m = model_.load();
                        auto s = std::ostringstream{};
                        {
-                           auto target = req.target().to_string();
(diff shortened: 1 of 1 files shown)


pinotree:
Note: using `std::to_string()` does not work for boost < 1.80 (i.e. with `boost::string_view`, and I suspect that is also why they switched to `boost::core::string_view`.

arximboldi (approved):
Thank you!

PR #198, opened 2024-01-02
Add an event loop that uses a lock-free queue

steve-mackinnon:
Hey! I'm considering using lager in a VST plugin where locking a mutex from the audio processing thread can be problematic. Using the `lock_free_queue_event_loop` proposed in this PR would allow thread-safe `context::dispatch()` calls without locking a mutex by relying on [moodycamel::concurrentqueue](https://github.com/cameron314/concurrentqueue) to shuttle event functions between threads in a lock-free manner.

Alternatively, I could shuttle actions over to the main thread to be dispatched in client code, but being able to call `dispatch()` safely from any thread seemed a bit more ergonomic. Curious to hear what you think. Thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,10 +81,16 @@ endif()
 if(lager_BUILD_TESTS)
   find_package(Catch2 REQUIRED)
 
+  set(concurrentqueue_INCLUDE_DIR "$ENV{CONCURRENTQUEUE_INCLUDE_DIR}")
+  if ("${concurrentqueue_INCLUDE_DIR}" STREQUAL "")
+    message(WARNING "Could not find concurrentqueue")
+  endif()
+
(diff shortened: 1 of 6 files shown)


steve-mackinnon:
Hey @arximboldi, thanks for checking this out! Great point regarding `std::function` allocating, I was overlooking that.

> In my own audio applications I use a different mechanism to send arbitrary closures or actions to the main thread or some other auxiliary non-RT thread, and do the dispatch from there. This method makes lock_free_queue_event_loop not needed.

Right on, that seems like the best approach. I'll do something similar for my use case.

Going to close out the PR. Thanks for the detailed response, much appreciated!

arximboldi:
My pleasure! I hope Lager ends up being useful for your use-case, curious about what comes out of it :)

PR #200, opened 2024-02-19
make match constexpr

asa:
Make lager::match constexpr.


asa:
This works for me on clang 17.0.6

arximboldi:
Thank you a lot Asa!

You can try running it in the `nix-shell` (see the parameters in the Github Actions log to reproduce the failures locally.

arximboldi:
Amazing! Merging!

PR #208, opened 2024-09-09
allow forward declaration of values held by reader

TheCoconutChef:
This commit enables the forward declaration of a type held by a reader without any new memory allocation or virtual methods. This is achieved by creating a new class  -- observable_reader_node -- in the reader_node hierarchy which acts as a view for the type held by reader_node.

Performance Impact

Ad-hoc benchmarks (i.e. not in commit) were performed with nanobench. (https://nanobench.ankerl.com/index.html)

Compared to baseline, the impact on a reader get method was non-existent. The same was true for map. The impact on watch seemingly ranged from indiscernible to rather small, i.e. in the order of 0.5-1.5%, although why there even is an impact isn't clear.

Downsides and Limitations

1. If it matters, the protected keyword has now been introduced within the reader node class hierarchy. This is because the link method and the observers access method had to be moved in observable_reader_node to avoid virtual method declaration.
2. While the values held by a reader are now fwd declarable, those held by a cursor are not. This arguably introduces an inconsistency in the API.

### DIFF
--- a/lager/detail/nodes.hpp
+++ b/lager/detail/nodes.hpp
@@ -128,26 +128,19 @@ struct notifying_guard_t
 };
 
 /*!
- * Base class for the various node types.  Provides basic
- * functionality for setting values and propagating them to children.
+ * Interface for nodes capable of notifying observers.
  */
 template <typename T>
-class reader_node : public reader_node_base
+class observable_reader_node
 {
 public:
     using value_type  = T;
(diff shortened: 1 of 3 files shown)


arximboldi (approved):
This is great, thank you!

arximboldi:
I agree that point 2. is somewhat an inconsistency, but we can work on this later.

PR #211, opened 2024-09-18
fix missing virtual dtor on observable_reader_node

TheCoconutChef:
This commit fixes an issue where the observable_reader_node, which serve as a base class for reader_node, didn't have a virtual dtor.

This was made easier to miss in virtue of observable_reader_node not having any virtual method.

By having observable_reader_node derive from reader_node_base, we effectively fix the issue and by removing multiple inheritance. observable_reader_node now inherits its virtual dtor from reader_node_base.


arximboldi:
Much better, yes, thank you!

PR #212, opened 2024-09-18
revert to index based notify loop

TheCoconutChef:
This fixes an issue in which calls to link occurring during notification would invalidate the iterators of the structured for loop. Calling link as a result of a reader<T>::watch callback can occur quite naturally.

By using the size of the children vector at the point at which the for loop is initiated, and by using the [] operator in order to access the children within the vector, the issue is avoided.

### DIFF
--- a/lager/detail/nodes.hpp
+++ b/lager/detail/nodes.hpp
@@ -235,8 +235,9 @@ public:
             bool garbage = false;
 
             this->observers()(last_);
-            for (auto& wchild : this->children()) {
-                if (auto child = wchild.lock()) {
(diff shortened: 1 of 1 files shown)


TheCoconutChef:
> Oh I missed this during the review! good catch! It would be nice to have a test to catch this one in the future.

Agreed. I'm trying my hand at a test for this. There's something tricky about the fact that the incorrect version of the loop falls into UB.

arximboldi:
Merging already, you can make a new PR for the test later if you feel like it...

Issue #215, opened 2024-11-11
force_opt combined with bind_opt results in compilation error


henri-v:
Creating a lens with force_opt as the innermost operation and then lifting this lens using bind_opt causes a compilation error. For example, adding the following code in test/type_erased_lens.cpp will not compile:

```.cpp
TEST_CASE("lenses, force_opt, type erasure, bind_opt")
{
    struct Foo { int i; };
    //const lager::lens<Foo, std::optional<int>> opt_member_lens =
    const auto opt_member_lens =
        attr(&Foo::i) | force_opt;
    const auto bound_opt_member = bind_opt(opt_member_lens);
    CHECK(view(bound_opt_member, std::optional{Foo{3}}) == std::optional{3});
    CHECK(view(bound_opt_member, std::optional<Foo>{}) == std::nullopt);
}
```

[Error message from GCC 11.4.0](https://gist.github.com/henri-v/fec6ca7df03a034ede7b0ae19fa53ab1)

By turning the force_opt lens into a type-erased lens, the code will compile, so there should be nothing wrong with the logic of the code. This can be used as a workaround.

It is possible to get the same result with map_opt. However, in some cases it’s better for code organisation that these steps are in separate functions, which can be combined more freely.

arximboldi:
Ping @Siapran

arximboldi:
Whatever you chose @Siapran sounds good, I trust you on this and would very much appreciate a PR if you have the time!

Siapran:
I'll try to do that today if whatever I caught on the flight back home will let me 👍

Siapran:
... I really should just rewrite the whole thing sooner than later. there's some really dubious (if not straight up wrong) moves going on in there.

arximboldi:
Fixed: https://github.com/arximboldi/lager/pull/216

Thank you so much @Siapran !

PR #216, opened 2024-11-13
Enable writing to force_opt lens with non-opt value.

Siapran:
fix for #215.

also fixes a couple potential use-after-move issues.


arximboldi (approved):
Great fix, thank you so much @Siapran ! That was super quick!

PR #217, opened 2025-03-27
Add with_thunk enhancer for thunk-style dispatch

demian51:
This adds a new enhancer called `with_thunk`, which enables dispatching not only plain actions but also thunk functions that take a `lager::context`.

This allows deferring logic such as conditional dispatch or side effects until runtime, similar to what redux-thunk provides in Redux.

Example usage:
```C++
store.dispatch([](lager::context<action> ctx) {
    // perform some effectful logic, then dispatch
    ctx.dispatch(some_action{});
});
```

### DIFF
--- a/lager/extra/thunk.hpp
+++ b/lager/extra/thunk.hpp
@@ -1,0 +1,85 @@
+//
+// lager - library for functional interactive c++ programs
+// Copyright (C) 2017 Juan Pedro Bolivar Puente
+//
+// This file is part of lager.
+//
+// lager is free software: you can redistribute it and/or modify
+// it under the terms of the MIT License, as detailed in the LICENSE
+// file located at the root of this source code distribution,
+// or here: <https://github.com/arximboldi/lager/blob/master/LICENSE>
+//
+
+#pragma once
+
+#include <lager/effect.hpp>
+#include <lager/util.hpp>
+
+#include <functional>
+#include <type_traits>
+#include <utility>
+#include <variant>
+
+namespace lager {
+
+/*!
+ * Enables dispatching thunk-style actions (functions that take a
+ * lager::context).
+ *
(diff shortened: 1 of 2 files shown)


arximboldi (changes requested):
Thank you a lot for this contribution!

It looks interesting, I'd be happy to merge it. Can you add a comment to `with_thunk` explaining this enhancer a little bit more?

Some questions:
1. I wonder, how does this affect the covariance that is allowed between action types, i.e. normally one can convert `context<A>` to `context<B>` as long as `B` is a subset of actions of `A` ?

2. Also, just out of curiosity, what is the use-case that motivated you to implement this? And why was using `ctx.loop().post(...)` not a viable alternative?

demian51:
Thanks for the feedback!

1. I believe `with_thunk` does not affect it. The context is passed through unchanged, so conversions like `context<A>` to `context<B>` should still work as before.

2. It was mostly curiosity. While `ctx.loop().post(...)` can handle similar cases, it requires capturing the context manually. With `with_thunk`, the context is passed in, so it's easier to reuse named functions and keep the dispatch flow a bit more explicit. That said, I agree it might be overkill for simpler use cases.

Also, I have been thinking that this could make it possible to build something like Redux Toolkit's `createAsyncThunk` on top, though I'm not sure yet how useful that would really be.

I'll add a brief comment to `with_thunk` as suggested.

Let me know if you see any issues or if there's a better direction to consider.

arximboldi:
Cool. Thank you for the clarifications. One last nitpick before merging, can you make the comment have this form, so that it can be referenced in the HTML docs later on:
```
/*!
 * The comment...
 */
```

demian51:
Sure! I've updated the comment to follow the Doxygen format.

Thanks for the thorough review!

PR #227, opened 2025-12-15
Switch from boost::asio::io_service to io_context

pinotree:
`boost::asio::io_context` has been available for many asio/boost versions, and `io_service` is a deprecated alias for it.

Since `io_service` and its bits (e.g. `io_context.hpp`) are being removed, switch to the proper name. There should be no behaviour changes.


pinotree:
This was included as part of commit f61a22f4829067a38676524b915776afc6a522ec.

Issue #236, opened 2026-03-21
Spurious notification on xform bind


BenLeadbetter:
[This bugfix patch](https://github.com/arximboldi/lager/pull/226) introduced a regression 🙈. After this change a `lager::bind` call to an xform reader node no longer correctly cashes the initial value and as a result we get one extra spurious notification when the state next changes:

```cpp
auto state = make_state(int{2}, automatic_tag{});
lager::reader<bool> even = state.xform(zug::map([](int v) {
    return v % 2 == 0;
}));

int count = 0;
even.bind([&count](auto) { ++count; });
// we expect one callback from the bind
REQUIRE(count == 1);

// this change doesn't cause a change in value for the reader - we don't expect a callback
state.set(4);
CHECK(count == 1); // fails! count == 2
```

PR #237, opened 2026-03-21
#236 | Bugfix | Spurious notification on xform bind

BenLeadbetter:
Contex: [#236](https://github.com/arximboldi/lager/issues/236)

### DIFF
--- a/lager/detail/nodes.hpp
+++ b/lager/detail/nodes.hpp
@@ -217,7 +217,6 @@ public:
             last_            = current_;
             needs_send_down_ = false;
             needs_notify_    = true;
-            has_sent_down_   = true;
             
             const auto& children = this->children();
             const size_t n_children = children.size();
@@ -270,7 +269,6 @@ protected:
     bool needs_send_down_ = false;
     bool needs_notify_    = false;
     bool notifying_       = false;
-    bool has_sent_down_   = false;
 };
 
 /*!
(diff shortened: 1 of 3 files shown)


lager/detail/nodes.hpp
```diff
@@ -217,11 +217,10 @@ class reader_node : public observable_reader_node<T>
             last_            = current_;
             needs_send_down_ = false;
             needs_notify_    = true;
-            has_sent_down_   = true;
```
BenLeadbetter, line 220: Caching the `has_sent_down_` state on `reader_node::send_down` was the wrong place because there are other entry points which would cause an `inner_node` to be evaluated before having been sent down - like in a `lager::bind` call!

lager/detail/xform_nodes.hpp
```diff
@@ -44,18 +46,26 @@ ZUG_INLINE_CONSTEXPR struct send_down_rf_t
...
+        return {std::apply(
+                    [&](auto&&... ps) {
+                        return xform(zug::last)(detail::no_value{}, ps->current()...);
+                    },
+                    parents),
+                false};
     } LAGER_CATCH(const no_value_error&) {
         if constexpr (std::is_default_constructible<ValueT>::value) {
-            return ValueT{};
+            return {ValueT{}, true};
```
BenLeadbetter, line 68: The source of data flow which causes a node to "not send down" is actually here. In this patch we pass this information on to the `inner_node` that the value was defaulted for whatever reason - perhaps filtering by an xform node.

arximboldi (approved):
Great catch and work investigateing this! too bad I missed this on the last review.

PR #238, opened 2026-07-29
Bugfix | MSVC warning | Fewer args than expected to macro function call

BenLeadbetter:
Clang and GCC will allow macro function calls to omit arguments, however [MSVC will emit a warning by default](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4003?view=msvc-170).

This small patch patch introduces no logical change, swapping the no-args call to `LAGER_THROW()` with its equivalent counterpart `LAGER_RETHROW`.


arximboldi:
Thank you!

